PR #25, opened 2018-04-02
JS code change with Google guide recommendations

domProjects:
change var by let, block indentation +2 spaces and more

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)


beeman:
The code in the dist folder is generated from the `src`, so probably very little use in formatting it.

With that being said, the best approach would probably be adding an eslint step after the build step and that way automate it :)

beeman:
Also, it's probably good to make several PR's for different changes in functionality :) 

Don't get me wrong, just trying to be helpful here 👍🏼

PR #57, opened 2018-04-05
Add variables for asset paths

beeman:
## Description

This PR adds some variables so the path to the assets can be set when using the scss in another project. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/src/assets/scss/_variables.scss
+++ b/src/assets/scss/_variables.scss
@@ -1,3 +1,5 @@
+$tabler-base: "../" !default;
+
 // Options
 $enable-rounded: true;
 $enable-shadows: false;
--- a/src/assets/scss/dashboard/_browser.scss
+++ b/src/assets/scss/dashboard/_browser.scss
@@ -1,3 +1,4 @@
+$browsers-path: '#{$tabler-base}/images/browsers/' !default;
 $browsers: (
 	'android-browser', 'aol-explorer', 'blackberry', 'camino', 'chrome', 'chromium', 'dolphin', 'edge',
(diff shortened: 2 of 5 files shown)


ollym:
I don't think you should do it this way.

You should have a function like `tblr_asset_path` which resolves the asset path. You could then override this function so this can be used with a package manager for example Sprockets in Rails or provide a custom base path.

beeman:
Hey @ollym, thanks for your input. I've simply copied how settings paths works in FontAwesome.

Feel free to send a PR with how you think it can be implemented easier.  

I would like to aim for the simplest implementation possible. Can you show an example of how your suggestion would work?

digilist:
I would really like to see this feature merged.

@ollym can you provide an example? I've never seen an asset path configuration with a function, only with variables like @beeman did it.

yatsyk:
> Hi guys, any way to merge it? It's impossible to add tabler to existing project via SCSS :(

As this pull request is not merged yet you can use tabler from fixed repo:

`npm install https://github.com/beeman/tabler.git#beeman/variable-asset-path --save`

```
//style.scss in your project
$tabler-base: "../node_modules/tabler-ui/src/assets";
@import '../node_modules/tabler-ui/src/assets/scss/variables';

//your changes 

@import '../node_modules/tabler-ui/src/assets/scss/dashboard/bootstrap/bootstrap';
@import '../node_modules/tabler-ui/src/assets/scss/dashboard/dashboard';
```

PR #66, opened 2018-04-06
layout fix in index.html and scroll fix issue in login and similar pages

sushant1221:
layout break in index.html : fixed
scroll removed from login and other similar pages like forgot password  : fixed


codecalm:
sorry, but `dist` directory are generated automatically, and should't be edited

sushant1221:
What about CSS. Can you  send me any link so that i can execute the project and generate dist from src

PR #102, opened 2018-04-17
"dimmer-content" opacity too much.

johnunclesam:
"dimmer-content" opacity too much.


src/assets/scss/dashboard/_progress.scss
```diff
@@ -93,7 +93,7 @@ Dimmer
 		}
 
 		.dimmer-content {
-			opacity: .4;
```
Prathamesh555: it should be 3 what say.!

PR #121, opened 2018-04-27
Remove display:block from .form-group

### LINKED ISSUE #111: Form Horizontal is not working
Hi,

form horizontal does not work.  ##

Greetings Serverjunge

chrisbaswell:
Removing this doesn't appear to affect the form layout but it does fix #111

### DIFF
--- a/src/assets/scss/dashboard/_form.scss
+++ b/src/assets/scss/dashboard/_form.scss
@@ -4,11 +4,6 @@ textarea {
 	}
 }
 
-
-.form-group {
-	display: block;
-}
-
 .form-label {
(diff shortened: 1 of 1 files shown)


ndbroadbent:
Thanks @chrisbaswell, I hope this gets merged! Just wasted a bit of time trying to figure out why the [horizontal form group example](https://getbootstrap.com/docs/4.1/components/forms/#horizontal-form) wasn't working (from the bootstrap 4.x docs.)

Here's the workaround I'm using for now:

```css
.form-group.row {
  display: flex;
}
```
For now

PR #126, opened 2018-05-03
Fix issue with broken 'tabler.svg' logo on 'Login', 'Register' and 'Forgot Password' pages

ptsurbeleu:
Noticed a broken Tabler logo on **Login**, **Register** and **Forgot Password** pages while I was browsing the theme.

The fix seems to be trivial, so I decided to address that since it didn't look right to show the demo with broken images.

Hope this would help to make things even better! 😃 

Please find below a comparison `before` vs. `after` for each page adjusted.

## Login
`before`
image
`after`
image

## Register
`before`
image
`after`
image

## Forgot Password
`before`
image
`after`
image

### DIFF
--- a/dist/forgot-password.html
+++ b/dist/forgot-password.html
@@ -43,7 +43,7 @@
           <div class="row">
             <div class="col col-login mx-auto">
               <div class="text-center mb-6">
-                <img src="./assets/brand/tabler.svg" class="h-6" alt="">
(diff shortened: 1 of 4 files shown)


ptsurbeleu:
@codecalm, let me know if the change is good enough to be accepted as is or you would need to create a work item for that.

Thanks!

codecalm:
thanks for contributing!

PR #132, opened 2018-05-06
Fix error on serving

mohe2015:
This fixes the following error: You have already activated public_suffix 3.0.2, but your Gemfile requires public_suffix 3.0.0. Prepending `bundle exec` to your command may solve this.


mohe2015:
You may want to update public_suffix to 3.0.2 in the Gemfile.

PR #168, opened 2018-07-18
Sketch source with all screens and controls

milosmatic:


krizanteme:
Can you share the vector format please, (recommend .sketch). Thanks!

Issue #79, opened 2018-04-12
Consider declaring the variables inside _variables.scss with !default


gnpretorius:
This allows the developer to create a custom variables file to tweak some of the basic properties. 

Example: `$font-family-base: 'Source Sans Pro', -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Arial, sans-serif !default;`

The custom file would be included before the `bundle.scss` file. 

If there is a better / different way of doing it, let me know
Cheers

codecalm:
okay, I will try to fix it quickly. Thanks for feedback!

simonmilz:
+1

tanshiqi:
+1

PR #230, opened 2018-11-25
Declare sass variables as default to allow overrides

bravo-kernel:
This PR implements the functionality requested at #79 and is totally backwards-compatible/non-obtrusive as it changes nothing (except that it now allows overriding these variables).

### Example: 
To change the main background color to white now simply set this **before** importing `dashboard.scss`:

```sass
$body-bg: #fff;
```


dsznajder:
@bravo-kernel Thank you for PR! This functionality is currently applied to `v1`

PR #239, opened 2019-01-10
Fix select-tag background SVG in Firefox

### LINKED ISSUE #117: Embedded SVG backgrounds don't work in Firefox
Classes such as `selectize-control.single` and `custom-select` that use an SVG embedded background to render their dropdown indicator fail when rendered on Firefox. This is due to Firefox reading instances of `#` in the data-uris as the start of an anchor section and the end of the URI.

Replacing `#` with `%23` I found to be an acceptable work around, although for some reason it breaks the automatic insert of `$custom-select-indicator-color` in to `$custom-select-indicator` in the variables file, so I manually insert that color for that instruction.

workgena:
[fixes #117]

Issue reported by @JamesLF #117

> Classes such as `.selectize-control.single` and `.custom-select` that use an SVG embedded background to render their dropdown indicator fail when rendered on Firefox. This is due to Firefox reading instances of `#` in the data-uris as the start of an anchor section and the end of the URI.

> Replacing `#` with `%23` I found to be an acceptable work around, although for some reason it breaks the automatic insert of `$custom-select-indicator-color` in to `$custom-select-indicator` in the variables file, so I manually insert that color for that instruction.

Issue appears only in Firefox. Fix works in Chrome as expected.

| Before   |      After      |
|----------|:-------------:|
| image |  image |

### DIFF
--- a/src/assets/scss/_variables.scss
+++ b/src/assets/scss/_variables.scss
@@ -270,7 +270,7 @@ $custom-select-padding-y: .5rem;
 $custom-select-padding-x: .75rem;
 
 $custom-select-indicator-color: #999;
-$custom-select-indicator: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 10 5'%3E%3Cpath fill='#{$custom-select-indicator-color}' d='M0 0L10 0L5 5L0 0'/%3E%3C/svg%3E");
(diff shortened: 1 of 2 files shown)


workgena:
Discussion on [StackOverflow](https://stackoverflow.com/questions/13367868/modify-svg-fill-color-when-being-served-as-background-image#comment61860173_27827090) about SVG as background-image:

> Note that you must urlencode the `#` character for your hex colors to make this work in Firefox. So something like `<svg fill="#ffffff" ...></svg>` becomes `<svg fill="%23ffffff" ...></svg>`

workgena:
After the #249 has been merged, this PR aims to fix regular `<select>` tag.

PR #245, opened 2019-01-27
text-decoration-skip fix

clarkeash:
The `text-decoration-skip` spec has changed


paulftw:
Went to github today to propose this change, but it's already under review :)

@codecalm please merge this PR :beer:

PR #265, opened 2019-03-25
Add menu items and fix _config.yml

WinterSilence:


dsznajder:
@WinterSilence Hi! thanks for PR but currently we are working on `v1` version so this is unrelated to next version.

PR #275, opened 2019-04-15
require jquery in core.js

level39:
$ is being accessed without requiring jquery which results in a $ is not a function.


dsznajder:
@level39 Thanks for PR! We are currently working on `v1` version which does not include require.js in it. Stay tuned for newest version

PR #283, opened 2019-05-12
Adds primary support for y2 axis

lejacobroy:
See 'Wind speed during two days' graph for a representation.

When two datasets differ greatly in amplitude, we often times need a second y axis to better represent their values.

Of course, we need to datasets (series) in the graph object, the first one being the y (left) and second one the y2 (right)


lejacobroy:
Here's two example : [Images](https://imgur.com/a/qyjm64n)

PR #293, opened 2019-06-18
Tabler docs

mrszympek:


dsznajder:
@codecalm I think we can merge it :)

PR #298, opened 2019-06-29
Build all packages separatelly

### LINKED ISSUE #296: SCSS plugins compilation
We need to compile tabler core (`tabler.scss`) and all tabler scss plugins like `tabler-*.scss`. 

All files like `.min.css`, `.min.css.map`, `css.map`, etc. also should be generated.

dsznajder:
closes #296


codecalm:
I've made scipts for custom plugins like `taber-flags`, `tabler-charts`, etc

Issue #300, opened 2019-07-06
Table sorting asc  or desc


electrocoder:
Table is default sorting. I change is date column is sorting asc  or desc. How can I do it?

WinterSilence:
@electrocoder https://datatables.net/extensions/rowreorder/examples/initialisation/simple.html

electrocoder:
Thanks

Issue #282, opened 2019-04-30
datatables extensions


Hanysamy25:
I can't add datatables extensions like print buttons please how to configure it

Tairau:
https://datatables.net/extensions/buttons/examples/print/simple

ViKingQ:
I need the datatables extensions like 'datatables.responsive', but I don't know how to configure it via requirejs

ViKingQ:
> https://datatables.net/extensions/buttons/examples/print/simple

How to configue it via requirejs, please

PR #303, opened 2019-07-09
Fix table sorting

WinterSilence:
Fix position of sort buttons in last th + format code style


dsznajder:
@WinterSilence Thanks but master is kinda outdated and I'm not sure if we can release version with your changes

CC @codecalm

WinterSilence:
@dsznajder #282  #300

dsznajder:
CC @mrszympek

Issue #219, opened 2018-10-27
Working with input files inside a form!


JustJunaid:
So, when i upload a file using tabler classes, the name of the file is not displayed, like for example the label on uploading 'data.csv' file does not change and remain 'Choose File', i guess it's related to the css, but i'm not sure. Can someone help me regarding this?

manuponce1993:
Hi, I was looking for a template to include in a project from my university and luckily I found this open source development. 
I think the problem is that there's no JS function that loads the name of the file attached to the input. Adding this function will do the magic!

```javascript
$(".custom-file-input").on('change',function() { 
	// Get the complete path and keep only the filename
        var filename = $(this).val().split('\\').pop();
        // Assign the filename to the associated label
	$(this).siblings(".custom-file-label").text(filename);
})
```

FreexD:
@JustJunaid @manuponce1993 it's been a very long time :D but it now works natively in the browser - no need to add custom JS for handling that ;) Check out tabler 1.0.0-alpha.15 for it ;)

PR #310, opened 2019-07-18
Update tabler.js

manuponce1993:
Hi, I was looking for a template to include in a project from my university and luckily I found this open source development.
I think the problem is that there's no JS function that loads the name of the file attached to the input. Adding this function will do the magic!

Issue reference: #219

### DIFF
--- a/js/tabler.js
+++ b/js/tabler.js
@@ -112,6 +112,13 @@ $(document).ready(function() {
     e.preventDefault();
     return false;
   });
+  
+  $(".custom-file-input").on('change',function() { 
+	      // Get the complete path and keep only the filename
+        var filename = $(this).val().split('\\').pop();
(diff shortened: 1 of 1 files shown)


js/tabler.js
```diff
@@ -112,6 +112,13 @@ $(document).ready(function() {
     e.preventDefault();
     return false;
   });
+  
+  $(".custom-file-input").on('change',function() { 
+	      // Get the complete path and keep only the filename
+        var filename = $(this).val().split('\\').pop();
```
dsznajder: Please use `const` as you do not modify this variable.

dsznajder: Also instead of `pop()`  use `.slice(0, -1)` as it does not modify source array.

PR #319, opened 2019-07-24
Update _config.yml

WinterSilence:
Add server settings for rewrite defaults. By default, `npm run start` opened `http://localhost:4567` for me. Also i switch on debug/profile modes - it's usefull in `dev`.


dsznajder:
@codecalm CC :)

PR #343, opened 2019-09-13
Update form-elements.html

WinterSilence:
JS moved to https://github.com/tabler/tabler/edit/dev/js/tabler.js


codecalm:
we moved `selectize` plugin to footer

PR #350, opened 2019-09-28
Changes for Bootstrap

kicaj:
I found old fix in bundle.scss: https://github.com/tabler/tabler/commit/8496732fd6d13067bba7bee5fda204aa338710a9
So now, it's problem with install tabler-ui by Webpack.

Path to Bootstrap "../../../bootstrap/scss/bootstrap.scss" doesn't work now.

I tested version with "~bootstrap/scss/bootstrap.scss" and "bootstrap/scss/bootstrap.scss" and works perfectly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "url": "https://github.com/tabler/tabler/issues"
   },
   "devDependencies": {
-    "bootstrap": "^4.1.0",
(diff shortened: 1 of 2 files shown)


codecalm:
I've fixed id in e2a15f7

PR #358, opened 2019-10-16
Fix selectize height issue

alex-safian:
Multi-select doesn't have a responsive height.

height issue

### DIFF
--- a/dist/assets/css/dashboard.css
+++ b/dist/assets/css/dashboard.css
@@ -19028,6 +19028,11 @@ svg {
   border: 0;
 }
 
+.selectize-control.multi {
+  min-height: 2.375rem;
+  height: auto;
+}
+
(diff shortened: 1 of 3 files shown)


dist/assets/css/dashboard.css
```diff
@@ -19028,6 +19028,11 @@ svg {
   border: 0;
 }
 
+.selectize-control.multi {
```
WinterSilence: fix [sass](https://github.com/tabler/tabler/tree/dev/scss) not css

alex-safian: Hey @WinterSilence, thank you for your review. Actually, I added my changes to the SASS file. I run `dist` command (`npm run dist`) to compile SASS files to the CSS files. Shouldn't I compile my changes?

PR #369, opened 2019-10-31
Fix invoice table jumping

alex-safian:
### Issue:
When a user clicks on the `action` button in the invoice table, dropdown makes some extra spaces below the table which causes jumping.

invoice-table-issue

### Solution
Adding `overflow-hidden` to the `table` element.


kicaj:
I think the problem is still present, like on homepage.

PR #384, opened 2019-11-13
mention Liquid so new users have correct expectations

gideonaa:
Maybe this is obvious to other people but as someone who doesn't do much front-end development, it took me some time to realize that this project uses Liquid. Though not a big change, I think this edit makes the docs more honest and will save others time.


codecalm:
thank you!

Issue #258, opened 2019-03-06
card-aside-column not responsive


enzorre:
Hi,

I'm noticing that the rendering of card-aside-column in the mobile is terribile because the image is not moved on the top of the card, but remaind on the left woth a small slice of window:

![image](https://user-images.githubusercontent.com/48055312/53867438-23ff1300-3ff4-11e9-898c-db1925a0f78e.png)

Are you agree? Please let me know how to solve, thanks.

MrNicodemuz:
The current preview documentation also seems to still have this issue:
https://preview-dev.tabler.io/cards.html

Does not render well on mobile screens.

<img width="566" alt="Screenshot 2020-03-22 at 11 30 28 PM" src="https://user-images.githubusercontent.com/1620454/77253439-326aac00-6c95-11ea-8c0f-047a13bf43f6.png">

codecalm:
Fixed:

<img width="484" alt="Screenshot 2020-03-29 at 19 14 49" src="https://user-images.githubusercontent.com/1282324/77855594-9fc89080-71f1-11ea-9830-8a8cdb4fbfa5.png">

PR #429, opened 2019-12-18
Fix: .card-aside and .card-aside-column reponsive

mendozagioo:
This solve issue #258

### DIFF
--- a/pages/_data/menu.yml
+++ b/pages/_data/menu.yml
@@ -12,32 +12,6 @@ forms:
   badge: New
   badge-color: red
 
-error:
-  title: Error pages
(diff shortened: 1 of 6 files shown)


codecalm:
hi @mendozagioo, in newest Tabler we removed `.card-aside` and we use bootstrap grid with 0 width gutters. See demo here: https://preview-dev.tabler.io/cards.html

mendozagioo:
Hi @codecalm, I thought so, but since the issue was still there, that's why I did the PR, could you close the issues that are already resolved to help in whatever need it? Please

mendozagioo:
I want to help in this project

PR #473, opened 2020-02-02
Fix users and invoices table jumping

alex-safian:
Fixing the tables jumping when the dropdown menu is opened. The default dropdown **boundary** option is `scrollParent` which affects the height of its parents so I change **boundary** option to `viewport` to prevent the jumping.


codecalm:
thank you for your contributing!

PR #626, opened 2020-05-04
Remove vertical-align for alert anchor elements

gryevns:
There is an issue with line alignment when you define a modal, with an alert & anchor, example: https://jsfiddle.net/57h8qdya/

Before
Screenshot 2020-05-04 at 19 25 33

After
Screenshot 2020-05-04 at 19 26 07


codecalm:
thank you for help! :)

PR #627, opened 2020-05-08
Avatar status badges should match the avatar size

danswiser:
This updates the status badge of avatar icons so that the badge will match the avatar size.

Before:
Avatar Status Sizes - Original

After:
Avatar Status Sizes - New

I believe the easiest option is to just set the width/height of the status badge to 1/4th the avatar w/h. From the options I tried, it looks the best while requiring hardly any changes.


codecalm:
thank you for your help!! :)

PR #663, opened 2020-07-15
🔥 Removed duplicate badges

jaredsmithse:
There were two duplicate OpenCollective badges, and after pulling apart the `<a />` tags I realized the HTML was invalid. I converted them to just be markdown images for easier readability as well 🎨


jaredsmithse:
Closing this, looks like @rajan-31 managed to close this out in #681

PR #705, opened 2020-11-03
Change the way of doing child process

Altffenser:
To build tabler successfully it is necessary to change the way a child process is carried out by making use of the cross-spaw package as shown.

1. Install `npm install cross-spawn`
2. Declare the variable as spawn
3. change on gulpfile.js, line 220 and 233

```
return spawn('bundle', ['exec', 'jekyll', 'build', '--watch', '--destination', demoDir], { stdio: 'inherit' })
		.on('close', cb);
```

```
return spawn('bundle', ['exec', 'jekyll', 'build', '--destination', demoDir], { env: env, stdio: 'inherit' })
		.on('close', cb);
```

The error displayed was:
`Error: spawn bundle ENOENT`

The error no longer exists thanks to @arthurattwell who explained the solution at this [stackoverflow post](https://stackoverflow.com/a/52233810/7972386)


codecalm:
hi @Altffenser! 

thank you for your PR, i've added your changes in https://github.com/tabler/tabler/commit/8be915e1e9c9eeabb65b60a2bcb7c267054866b0 cause here are a lot of conflicts, 

thank you for your help! :)

PR #737, opened 2020-12-12
Add border width control to card class

enricodias:
The variable ```$card-border-width``` was created dynamically with the default value of ```1px``` but it was never used define cards borders. I added it to the card class allowing it to actually control the borders.


codecalm:
<3 thank you! :)

PR #739, opened 2020-12-13
Improve cards with no border

enricodias:
Changes:

- ```.border-0``` is applied to all child cards and tabs. To build a page with lots of borderless cards this class can be set on the parent element instead of on each individual card.
- Prevents ```.card-active``` from breaking like [this](https://i.imgur.com/TkLge82.png).
- Reduces the opacity of ```.card-stacked``` when ```.border-0``` is active to keep the stacked look.
- Adds new examples in the cards page.


codecalm:
thank you! <3

PR #770, opened 2021-01-06
Some dark theme fixes

kilab:
I found a few more bugs in dark mode so I fixed for myself and I share effects my work. 

There are more things to improve but I can't fix everything at once. I develop private project which using tabler so you can expect more small fixes added occasionally.


codecalm:
thank you for help! ❤️

PR #779, opened 2021-01-18
Make lite validation also work properly for form select

rjd22:
Screenshot from 2021-01-18 12-28-02

Lite validation was only work for form fields with the `.form-control` class. Selects
use the `.form-select` class. This change also applies lite validation for the selects.


rjd22:
@codecalm this is a very small change. Could you take a look at it?

codecalm:
❤️   thank you for your work! :)

PR #788, opened 2021-01-30
Use latest version of terser (^5.5.0)

charescape:
`gulp-minify` uses `terser 3.x` which is no longer maintained

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,7 +6,7 @@ const gulp = require('gulp'),
 	header = require('gulp-header'),
 	cleanCSS = require('gulp-clean-css'),
 	rtlcss = require('gulp-rtlcss'),
-	minifyJS = require('gulp-minify'),
+	child_process = require('child_process'),
 	rename = require('gulp-rename'),
 	purgecss = require('gulp-purgecss'),
 	rollupStream = require('@rollup/stream'),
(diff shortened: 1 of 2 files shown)


codecalm (approved):

codecalm:
I've check it and build crashed :/
what do you think about https://www.npmjs.com/package/gulp-terser ?

charescape:
@codecalm To avoid some troubles caused by 3rd-party plugins stop maintenance(in some day), I think `terser` command line should be used directly instead of using gulp-plugin, like `twbs/bootstrap` https://github.com/twbs/bootstrap/blob/main/package.json#L46-L48

gulpfile.js
```diff
@@ -223,13 +223,7 @@ gulp.task('js', () => {
...
 
 	if (BUILD) {
-		g.pipe(minifyJS({
-			ext: {
-				src: '.js',
-				min: '.min.js'
-			},
-		}))
-			.pipe(gulp.dest(`${distDir}/js/`));
+		 child_process.execSync(`terser dist/js/tabler.js --compress passes=2 --mangle --output dist/js/tabler.min.js`, {encoding: "utf8"});
```
FreexD, line 226: @charescape do you think you could fix this command to work on netlify as well?

PR #836, opened 2021-04-15
fix #746

### LINKED ISSUE #746: [BUG] Tables header has wrong color in darkmode
**Describe the bug**

Since the last update (Release 1.0.0-alpha.21 maybe already Release 1.0.0-alpha.20) the header of a table is no longer in dark mode, but in a kind of white _(Before and after screenshots below)_

**Device**

 - Browser: Chrome Version 87
 - OS: Windows 10 and Android 10
 - Screen size: All

**To reproduce**

1. Switch to `theme-dark`
2. Look at a Table (Example at: https://preview.tabler.io/docs/tables.html - Run `document.body.classList.add('theme-dark');` in the Browser Console)

**Screenshots**
Bevor:
image

Now:
image
image

brylocode:
headers work, but i fixed text color in `tbody`

### DIFF
--- a/src/scss/_dark.scss
+++ b/src/scss/_dark.scss
@@ -92,6 +92,10 @@
     background: transparent;
   }
 
+  .table tbody{
+    color: $dark-mode-text;
+  }
+
   .list-group-item,
   .form-fieldset {
     border-color: $input-border-color;


src/scss/_dark.scss
```diff
@@ -92,6 +92,10 @@
     background: transparent;
   }
 
+  .table tbody{
```
FreexD, line 95: please fix table-striped as well! 
image

PR #837, opened 2021-04-15
fix #785

### LINKED ISSUE #785: [BUG] theme-dark class overrides background colors on avatar class
**Describe the bug**

the `.theme-dark` class overrides the background-color of the `.avatar` elements that have a `.bg-<color>` class assigned to them.

**To reproduce**

1. Go to https://preview.tabler.io/docs/colors.html
2. Add class "theme-dark" to the body element
3. All colored avatar examples will have the same gray background

**Screenshots**

Screenshot (9)

brylocode:
fixed avatars colors in dark theme

### DIFF
--- a/src/scss/_dark.scss
+++ b/src/scss/_dark.scss
@@ -60,10 +60,6 @@
     background-color: $dark-mode-darken;
   }
 
-  .avatar {
-    background-color: $dark-mode-lighten;
-  }
-
   .avatar-list-stacked .avatar {
     box-shadow: 0 0 0 2px $dark;
   }


src/scss/_dark.scss
```diff
@@ -60,10 +60,6 @@
     background-color: $dark-mode-darken;
   }
 
-  .avatar {
```
FreexD, line 63: this is not the right way to fix that - default avatars now do not have dark mode at all
image

FreexD (changes requested):
Make the fix somewhat similiar to how lt avatars work in dark mode.

image

PR #924, opened 2021-10-27
[Suggestion] Move optional dependencies to peerDependencies

rjd22:
When requiring Tabler it will install all the dependencies like Apexcharts, Choices.js, etc... even when you don't really use them. This increases the dependency and packages graph by a lot. It's good practice to encourage users of Tabler to only select the packages they need so that they have more control on the dependencies that are in their `node_modules` folder.

It might be that we need to fix some parts in the CI to make sure `npm ci` is run in place of `npm install`. This will always install the peerDependencies for testing and building.

More info on `peerDependencies`: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependencies

IMPORTANT: This is a breaking change so don't merge right away. @codecalm what do you think about this change?


rjd22:
@codecalm I added documentation and this PR could be merged. But somehow the netlify deployment doesn't have styles anymore but I can't see what part I'm doing wrong in my PR.

codecalm:
I'll check it :)

rjd22:
@codecalm Fixed the issue. Are you okay if I merge this?

codecalm:
yes, you can merge it

PR #935, opened 2021-11-14
Chat component

codecalm:


kevinpapst:
I'd say: both chat partners can get the same bubble color in dark mode.
So in dark mode the class `.chat-message-bubble` could have a `background-color: transparent`. Current it is white on white and unreadable.

codecalm:
@kevinpapst it's a draft at this moment :)

kevinpapst:
Yeah, I assumed you knew that. But I use tabler in my app and plan to help with (at least) feedback in the future.

oranyero:
When will the chat option be available?

oranyero:
No update about the chat?

codecalm:
I've rebuild chat component here: https://github.com/tabler/tabler/pull/1636

PR #944, opened 2021-11-22
Fix text body not overruling dark mode properly

rjd22:


kevinpapst (approved):
I can confirm that I have a rule like this is my project to fix the issue:
```
@include dark-mode {
    /* list items */
    .text-body {
        color: $dark-mode-text!important;
    }
}
```
I'd appreciate if that gets fixed in tabler directly. There are some other issues with the dark mode, I'll post them for discussion in a new issue.

Issue #945, opened 2021-11-22
[BUG] Dark mode issues


kevinpapst:
**Describe the bug**

After integration of Tabler into a larger project I found several minor problems with the dark mode. Unfortunately I didn't add comments what exactly was the issue, but the rules are so simple that verification should be 
I fixed these issues with the following SASS:

```scss
@include dark-mode {
    .link-dark {
        color: $dark-mode-text;
    }
    .bg-gray, .bg-default {
        color: $dark;
    }
    pre {
        background-color: transparent;
    }
    .btn-close {
        color: $dark-mode-text;
    }
    /* list items */
    .text-body {
        color: $dark-mode-text!important;
    }
    .list-group-item {
        border-color: $border-color-dark-transparent;
    }
}
```

The text body will be fixed with #944 by @rjd22 

EDIT: if you tell me how you want to see PRs, I am happy to contribute them. I just know that it can be painful to get CSS changes as PRs, because it needs a full-rebuild on your end to review...

kevinpapst:
Ok, I'll work on the PRs. Thanks for your quick feedback @rjd22

kevinpapst:
`pre` rule before:
<img width="944" alt="Bildschirmfoto 2021-11-24 um 22 30 03" src="https://user-images.githubusercontent.com/533162/143315262-aee83843-714f-4620-9394-f45ff68df766.png">
And after:
<img width="886" alt="Bildschirmfoto 2021-11-24 um 22 30 33" src="https://user-images.githubusercontent.com/533162/143315273-7034b8a2-bc4f-4446-9d17-3f7a20b24e1e.png">

kevinpapst:
Regarding the `.btn-close` rule I was wrong, as I am using FontAwesome instead of Tabler icons.
The actual problem is the SVG used as background which has set 
```
transparent url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' fill='%23000000'%3e%3cpath d='M.293.293a1 1 0 011.414 0L8 6.586 14.293.293a1 1 0 111.414 1.414L9.414 8l6.293 6.293a1 1 0 01-1.414 1.414L8 9.414l-6.293 6.293a1 1 0 01-1.414-1.414L6.586 8 .293 1.707a1 1 0 010-1.414z'/%3e%3c/svg%3e") center/0.75rem auto no-repeat
```
but should be (check `fill` attribute)
```
[...]

Issue #897, opened 2021-07-16
[BUG] active tab in dark mode is white


dheineman:
**Describe the bug**

I am not sure if this is intentional but in case it is not here goes.

When using the dark theme for tabler the active tab has a bright white color

**Device**

 - Browser: Google Chrome Version 91.0.4472.114 (Official Build) (x86_64)
 - OS: MacOS 11.2.3
 - Screen size: 2560×1440

**To reproduce**

1. Go to https://preview.tabler.io/tabs.html
2. Open the element inspector and add the `theme-dark` class to the body

**Screenshots**

![Screenshot 2021-07-16 at 11 19 36](https://user-images.githubusercontent.com/516028/125924749-134b1114-87e2-46b6-b3c1-f3bb6a252d83.png)

kevinpapst:
@codecalm could you share some insights on this topic? Is it by intention that the tabs are the only (as far as I could see) elements with pure white borders and such a strong accent color or are you open for changes here?

codecalm:
I'll check it

rjd22:
This is fixed now

kevinpapst:
@rjd22 this is not fixed, see https://preview.tabler.io/tabs.html

<img width="491" alt="Bildschirmfoto 2022-03-24 um 10 53 56" src="https://user-images.githubusercontent.com/533162/159890275-82841b92-3ee1-471a-9fd5-d189cb9a10d3.png">

Or is this really intentional?

PR #950, opened 2021-12-02
Dark mode improvements

codecalm:
demo: https://deploy-preview-950--tabler.netlify.app/?theme=dark

localhost_3000_index html
localhost_3000_wizard html
localhost_3000_layout-vertical-transparent html
localhost_3000_modals html
localhost_3000_search-results html
screencapture-localhost-3000-modals-html-2021-12-03-00_12_23
localhost_3000_music html
localhost_3000_charts html
localhost_3000_form-elements html
localhost_3000_gallery html


codecalm:
from now, we can enable dark mode in navbar:

Screenshot 2021-12-03 at 00 07 17

kevinpapst:
That is a great addition, finally no need to use the developer console to change the theme on each demo page :)

Can you share the why you changed from `class="theme-dark"` to `body[data-theme="dark"]`? I don't fully understand the change, but it feels like a big BC break and (without known any background about this change) feels wrong to me. Is this only for the demo? Why did you choose to use a data-attribute instead of a class name? I see the code being used in the demo JS but also some of the the SASS files.

Some points I found: 
- `<pre>` has still white background: https://deploy-preview-950--tabler.netlify.app/docs/form-selectboxes.html
- Same page "Advanced selectboxes": the selected title gets dark and is not visible
- the [highlighted row](https://deploy-preview-950--tabler.netlify.app/tables.html) in `table-striped` should probably blend in more instead of being plain white
- Is it desirable that a [dark button](https://deploy-preview-950--tabler.netlify.app/playground-dark.html?theme=dark) won't be shown in dark theme?

At least the demo has huge differences between modals in the [documentation](https://deploy-preview-950--tabler.netlify.app/docs/modals.html) and on the [modals page](https://deploy-preview-950--tabler.netlify.app/modals.html). Some elements are not shown. It seems the same happens to other elements (see "Advanced selectboxes" above).

Maybe you can have a look at https://github.com/tabler/tabler/issues/945 and include the fixes?

rjd22:
@codecalm looks good to me for a initial version. A few recommendations:

* I would normally use cookies for storing dark mode. This way backend systems can also know if dark mode is enabled
* Is there a technical reason you switched from the class to a data attribute? I prefer a class because it's way simpler to toggle.

I actually made a similar implementation in vue that toggles the dark mode without refreshing the whole screen. This way forms are remembered. I will share it as inspiration and remove the vuejs bits:

```js
    let darkModeEnabled = false;

    const enableDarkMode = (isEnabled) => {
      darkModeEnabled = isEnabled;

      document.body.classList.toggle('theme-dark');
      document.cookie = `darkMode=${isEnabled}; path=/; SameSite=Strict; Secure`;
    };

    const cookies = document.cookie.split(';');
    const userPrefersDarkMode = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;

    if (cookies.includes('darkMode=true') || cookies.includes('darkMode=false')) {
      darkModeEnabled = cookies.includes('darkMode=true');
    } else if (userPrefersDarkMode) {
      enableDarkMode(true);
    }

    return {
      darkModeEnabled,
      enableDarkMode,
    };
```

src/pages/_layouts/base.html
```diff
@@ -39,7 +39,7 @@
 </head>
 
 {% assign layout-dark = page.layout-dark | default: site.layout-dark %}
-<body class="{% if layout-dark %} theme-dark{% endif %}{% if layout.body-class %} {{ layout.body-class }}{% endif %}{% if page.body-class %} {{ page.body-class }}{% endif %}">
+<body {% if layout.body-class or page.body-class %} class="{% if layout.body-class %} {{ layout.body-class }}{% endif %}{% if page.body-class %} {{ page.body-class }}{% endif %}"{% endif %} data-theme="light" data-menu-position="top" data-menu-behavior="fixed" data-container-layout="boxed">
```
rjd22, line 42: This is the only thing I'm unsure about. Is there a technical or another reason why classes would not work? I would say that a class is the right use for dark theme's.

rjd22 (changes requested):

src/pages/_layouts/base.html
codecalm: no problem, i'll change it

codecalm: fixed

codecalm:
@rjd22 its just for demo, not light/dark mode mechanism. `demo.js` will not be included in main package.

rjd22:
@codecalm lgtm. :shipit:

rjd22 (approved):

kevinpapst:
I would ask for the following fix, even if it wasn't part of the original PR intention:
- the highlighted row in table-striped should probably blend in more instead of being plain white

Some content (here: "Egan Poetz") is only visible, if you mark some text with the cursor: 
Bildschirmfoto 2021-12-03 um 15 05 56

Everything else looks great 👍

codecalm:
@kevinpapst fixed :)

kevinpapst:
Found one more: 
divs with `bg-light` as in [Borders](https://deploy-preview-950--tabler.netlify.app/docs/borders.html) and [Cursors](https://deploy-preview-950--tabler.netlify.app/docs/cursors.html) look weird. At least the font color should be adjusted. Or is that only an example content problem?

rjd22:
@codecalm I pushed a separate branch with chart fixes for dark mode to avoid conflicts. If you like you can cherry pick the change:

https://github.com/tabler/tabler/pull/952

kevinpapst:
A tiny one: I believe that `.nav-tabs` currently has no dark styling. 
https://deploy-preview-950--tabler.netlify.app/tabs.html
Seems that these borders are the only ones which are so light.

Seems this was mentioned in #897 before.

jaikt:
som how'd I implement the dark mode and light mode toggle and everything.?

PR #952, opened 2021-12-03
Fix apexchart tooltip styling in dark mode

rjd22:


codecalm:
i've merged it to #950

PR #956, opened 2021-12-08
Update _mixins.scss

migolo:
Fixes Error: Top-level selectors may not contain the parent selector "&".
As specified here https://github.com/sass/sass/issues/1873#issuecomment-152293725

### DIFF
--- a/src/scss/mixins/_mixins.scss
+++ b/src/scss/mixins/_mixins.scss
@@ -14,26 +14,26 @@
 }
 
 @mixin scrollbar {
-  &::-webkit-scrollbar {
+  #{if(&, "&", "*")}::-webkit-scrollbar {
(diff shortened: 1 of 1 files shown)


rjd22:
@codecalm most likely Chromatic is blocking PR's because external PR's have no access to credentials as a security measure. Maybe it's an idea to disable Chromatic for PR's?

kevinpapst:
A simple fix could be to adapt the workflow:
```
jobs:
  chromatic-deployment:
    runs-on: ubuntu-latest
    if: ${{ github.actor == 'codecalm' }}
```
Then it will only start if the PR was started by @codecalm and thus contains the secret.

rjd22:
@codecalm is it possible to take a look at this when you have time?

PR #965, opened 2021-12-13
New statuses

codecalm:
localhost_3000_docs_statuses html_theme=light


kevinpapst:
Nice ones 👍 

One question: couldn't you re-use [all existing bg-* colors classes](https://preview.tabler.io/docs/colors.html) instead of creating new ones?

Are PRs/branches auto-deployed somewhere so we can have a look without building locally?

codecalm:
@kevinpapst its not possible to use `bg-*` classes, because I need rgba color values :( 

you can view preview here: https://deploy-preview-965--tabler.netlify.app/ :)

kevinpapst:
All "normal" ones look exactly the same with the bg-* classes, only some of the animated ones look wrong when using bg-* classes. But yeah, you know what you are doing 😄 

Any idea why Indigo is wrapped on page load? Maybe this is a Safari bug, it doesn't happen on Firefox/Chrome. The spaces do not collapse as as well (which I have never seen before).
Bildschirmfoto 2021-12-17 um 01 32 48

rjd22 (approved):

rjd22:
@codecalm do you like this enough to merge it? Or is it unfinished?

codecalm:
@rjd22 I'll resolve conflicts and I'll merge it :)

PR #976, opened 2021-12-16
Replace `divide(...)` with `math.div` in the SCSS files

paulftw:
Honestly, I could not understand how `divide($foo, $bar)` works in `@tabler/core` but for some reason it doesn't work when I `@use "@tabler/core/.../_variables.scss"` in my project.

Based on the [SASS deprecation notice](https://sass-lang.com/documentation/breaking-changes/slash-div) `math.div` is the way to go regardless.
WDYT about this change?


src/pages/_plugins/jekyll-filters.rb
```diff
@@ -45,10 +45,6 @@ def first_letters(value)
       value.to_s.split.map(&:chr).join
     end
 
-    def divide(value, number)
```
codecalm, line 48: it's jekyll filter, not sass mizin, you cannot remove it, becaouse for example this line will be not work: 

```
{% assign all-days = days | minus: start-day | plus: 1 | divide: 7.0 | ceil | times: 7 %}
```

codecalm (changes requested):

codecalm:
`divide($foo, $bar)` is made in bootstrap core, Tabler just use it

rjd22:
@codecalm I reverted the divide removal. I'm not sure you want to merge this even then since it's build based on a misconception so I leave merging or closing to you :)

paulftw:
@codecalm  @rjd22  this branch had conflicts, and in the end I decided it was easier to start a new one (don't ask).

I've provided more detailed rationale there and no longer touch jekyll-filters

Please see #994

Issue #984, opened 2021-12-31
[BUG] Extra background colors not working on table/td elements


superfaz:
**Describe the bug**

The bg-color classes don't work when applied to a <td> element.

**Device**

 - Browser: All - confirmed on Edge and Firefox
 - OS: W11
 - Screen size: 3000x2000

**To reproduce**

1. Go to https://jsfiddle.net/5xyqaezf/
2. reference: The 'secondary' cell works fine
3. error: The 'orange' cell has no background

**Identified cause**

There is a conflict between <td> style - which define a background-color - and the extra bg-color classes - defining a background only.
More over the boostrap standard bg-color classes have an !important flag which is missing on the extra bg-color classes.

PR #985, opened 2021-12-31
Update _colors.scss - Fix for #984

superfaz:
Fix for https://github.com/tabler/tabler/issues/984 - [BUG] Extra background colors not working on table/td elements


rjd22:
Thank you for the fix. This will be in the next release.

PR #1027, opened 2022-02-07
add font-display: swap; to improve font loading

AnTheMaker:
I added the `font-display: swap;` tag when loading the Inter font to improve the experience on slower connections.
It will make the text appear immediately with a fallback font and swap it with the Inter font as soon as the font is loaded. This shouldn't have any visual change for most users with a proper internet connection, but for slower connections - especially on mobile - this can make a huge difference.
More infos: https://developers.google.com/web/updates/2016/02/font-display#swap

### DIFF
--- a/src/scss/fonts/_webfonts.scss
+++ b/src/scss/fonts/_webfonts.scss
@@ -5,6 +5,7 @@
   @font-face {
     font-family: $font-local;
     font-style: normal;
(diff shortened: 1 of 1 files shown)


codecalm:
thank you! I forgot about that

AnTheMaker:
Awesome, thank you! That was fast :)

PR #1042, opened 2022-02-17
Update accordion.html

SayaliPatiil:
duplication issue


AnTheMaker (commented):
This PR seems to be spam. The only thing changed is the capitalization of some (random) HTML tags...

Issue #1028, opened 2022-02-07
[BUG] Inter font rendering issue


kevinpapst:
**Describe the bug**

Since the integration of local font, I am seeing this issue for capital A & Q eg. [on this page](https://preview.tabler.io/form-elements.html): 

<img width="449" alt="Bildschirmfoto 2022-02-07 um 12 57 21" src="https://user-images.githubusercontent.com/533162/152783999-a5a43c87-d772-4fd4-805b-2bbc87a42684.png">

There is a [pretty long thread about this issue](https://github.com/google/fonts/issues/2602), so I guess you cannot fix anything.
Just wanted to raise awareness that the integration of the local font was a nice move, but caused problems as well...

This is pretty annoying in my app (German) where now 3 of 7 main menu items look strange:
<img width="892" alt="Bildschirmfoto 2022-02-07 um 13 02 34" src="https://user-images.githubusercontent.com/533162/152784609-30f06ca7-8393-4d97-8813-abcef6473dab.png">

**Device**

MacOS 10.15.7 with Safari 15.2 but also on Chrome 98.

AnTheMaker:
That seems like an interesting bug... does this only happen with the woff2 files sourced from google fonts or also with the original font files from https://github.com/rsms/inter?
Alternatively, we could also try to use the variable font from the rsms/inter repo which shouldn't have this bug (afaik).

Issue #1033, opened 2022-02-09
[BUG] Error: Can't resolve '../../node_modules/@tabler/core/src/scss/fonts/inter-v7-latin-ext_latin-300.woff2'


reina3596:
In a clean installation I receive with Laravel I receive

Error: Can't resolve '../../node_modules/@tabler/core/src/scss/fonts/inter-v7-latin-ext_latin-300.woff2'

PR #1048, opened 2022-02-19
Remove `Inter` font and use default font system stack

codecalm:
Using custom fonts was very problematic. Now we use the default font set. The user can add his own fonts and use them. I added in the documentation a description how to do it. Custom Google font can be used as CSS variable.

fixes: #1028, #1033


kevinpapst:
Thanks for taking care of the issue! It feels right to let the user decide which font to use. 

By any chance, could you add a chapter to the new docs page that explains how to set the font with custom SCSS? 
From looking at the code: can I set a variable like `$font-google: 'Inter';`? 
Oh and btw. the variable name `font-google` seems to be misleading, probably change it to something like `font-custom`.

rjd22:
@codecalm looks great 👍. Germany already banned the use of Google fonts so it's indeed better to get used to using system fonts.

rjd22 (approved):

PR #1049, opened 2022-02-20
Update markdown.html

diomed:
- added underline which was missing


codecalm:
The underline tag has been deprecated as of HTML4. The W3C reference can be found here. The reason is that visual styling does not belong in tags, but should be moved to style-sheets.

https://stackoverflow.com/questions/1078912/why-is-element-u-considered-outdated-according-to-vs2008

PR #1081, opened 2022-03-07
Revert active list item edge highlight

df2:
Keep it minimal

### DIFF
--- a/src/scss/ui/_lists.scss
+++ b/src/scss/ui/_lists.scss
@@ -26,7 +26,6 @@
 
 .list-group-item.active {
   background-color: inherit;
-  border-left: 2px solid $yellow !important;
 }
 
 .list-group-item {


rjd22:
@df2 can you share a screenshot on what this change does? That way it's easier for me to review.

df2:
Sure @rjd22. After comparing the two again, I like it better with the edge highlight. What do you think?

With edge highlight (current code):
Screenshot_20220324_195938

Without edge highlight:
Screenshot_20220324_195856

rjd22:
Yep, I would not remove the edge highlight. Is it okay to close this PR in that case?

df2:
Absolutely. Will close it with this comment. I sent in another pull request (#1100) with a refactor of the properties of the highlight css that that's just a cleaner way to do it.

PR #1091, opened 2022-03-23
Fix toasts in dark-mode being to bright

rjd22:
Before:
Screenshot from 2022-03-23 10-35-04
After
Screenshot from 2022-03-23 10-39-54


kevinpapst:
Oh yes, that is so much better 😄 Thanks for fixing 👍

PR #1100, opened 2022-03-25
Replace `!important` modifier with more specific selectors ...

df2:
… and also replace hard coded border width with `$border-width-wide` variable.

### DIFF
--- a/src/scss/ui/_lists.scss
+++ b/src/scss/ui/_lists.scss
@@ -26,7 +26,8 @@
 
 .list-group-item.active {
   background-color: inherit;
-  border-left: 2px solid $yellow !important;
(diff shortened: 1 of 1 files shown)


rjd22:
I'm leaving this one for @codecalm I don't mind but since this is more of a style change maybe @codecalm has a opinion in this.

df2:
Sounds good @rjd22. I haven't tested it but it should also make it easier for rtl styles to move the highlight to the right.

df2:
RTL style works fine with it 👍

Arabic example:
Screenshot_20220325_081255

codecalm:
thank you for help! :)

PR #1098, opened 2022-03-24
Fix #1024: Fix Tom-select in dark mode

rjd22:

Issue #1024, opened 2022-02-06
[BUG] Dark mode issue for Tomselect


kevinpapst:
**Describe the bug**

Some issues with Tomselect exist in dark mode at https://preview.tabler.io/form-elements.html:

- Selected text is not visible
- Not possible to identify optgroups (adding an example in one of the demo dropdowns would be great!) 

<img width="429" alt="Bildschirmfoto 2022-02-06 um 18 03 42" src="https://user-images.githubusercontent.com/533162/152692103-9aca8d35-378b-47f4-87d8-b1fc90b7e8bc.png">

Nejcc:
Why this dropdown is not full width?

rjd22:
@kevinpapst would it be possible to make an code example of the optgroups for me? I might fix that too.

rjd22:
Outgroups still need to be fixed.

kevinpapst:
Yeah sorry, didn't follow the bug. Main point is fixed. 

The optgroup is still white, but that is less problematic:

<img width="423" alt="Bildschirmfoto 2022-07-30 um 13 14 53" src="https://user-images.githubusercontent.com/533162/181908422-ca969377-e787-4244-a110-8cbf0ee49d3a.png">

rjd22:
If you're willing and have time to make an PR for it I will gladly merge it ;).

kevinpapst:
I'd be happy to invest more, but I am really missing the communication how the project is going forward. Is the dark mode replaced by the Bootstrap native dark mode that is now in alpha/beta? Is it even worth investing any time into the dark mode right now? 

How many breaking changes can we expect in the beta cycle (usually only used for bug fixing and not for introducing major changes) before a stable release? Not that this has anything to do with this issue right here, more of a general uncertainty that holds me back.

rjd22:
@codecalm would be the one to answer correctly but I would guess that the original dark mode will not be replaced with Bootstrap dark mode soon.

It would be an good idea to focus releasing a stable version and improving tabler on a feature by feature basis.

PR #1108, opened 2022-04-11
added optgroup form examples

kevinpapst:
@rjd22 here is a PR that adds optgroup examples in the form page, so they can be tested easily

See also #1024 and #1098

This is default browser (Safari) rendering:
Bildschirmfoto 2022-04-11 um 12 27 47

And this is how Tom-Select currently looks:
Bildschirmfoto 2022-04-11 um 12 28 41

This is how it looks normally with TomSelect and Bootstrap 5:  https://tom-select.js.org/examples/optgroups/

**UPDATE**
I found the issue that is causing it.
You are importing the [tom select bootstrap 5](https://github.com/tabler/tabler/blob/main/src/scss/vendor/_tom-select.scss#L1) css file, but you are preventing the rules from being applied. They are prefixed like this: `.#{$select-ns}-dropdown` where `select-ns` variable is `ts` by default, resulting in rules like these:
```css
.ts-dropdown {
	.optgroup-header {
		font-size: $font-size-sm;
		line-height: $line-height-base;
	}
}
```

But the classes are [not applied in the select.html include](https://github.com/tabler/tabler/blob/main/src/pages/_includes/ui/select.html#L59-L60).
Is that on purpose @codecalm? By re-applying them (I did in the PR) the dropdown now looks better:
Bildschirmfoto 2022-04-11 um 12 55 08

Now dark mode still has an issue, it shows the header with white background. This rule makes it slightly better:
```css
.ts-wrapper{
  .dropdown-menu {
    .optgroup-header {
      color: var(--#{$variable-prefix}body-color);
      background-color: var(--#{$variable-prefix}body-bg);
    }
  }
}
```
but I am not entirely happy with the result and would like to leave this up to you, because you simply know better.

### DIFF
--- a/src/pages/_includes/ui/select.html
+++ b/src/pages/_includes/ui/select.html
@@ -65,7 +65,7 @@
 		var el;
 		window.TomSelect && ({% if jekyll.environment == 'development' %}window.tabler_select["select-{{ id }}"] = {% endif %}new TomSelect(el = document.getElementById('select-{{ id }}'), {
 			copyClassesToDropdown: false,
-			dropdownClass: 'dropdown-menu',
+			dropdownClass: 'dropdown-menu ts-dropdown',
 			optionClass:'dropdown-item',
 
 			{% unless include.show-search %}
--- a/src/pages/_includes/ui/select.html
+++ b/src/pages/_includes/ui/select.html
@@ -20,6 +20,15 @@
 				<option value="{{ person.id }}"{% if include.indicator == 'avatar' %} data-custom-properties="{% capture indicator %}{% include ui/avatar.html size='xs' person=person %}{% endcapture %}{{ indicator | strip | escape }}"{% endif %}>{{ person.full_name }}</option>
 			{% endfor %}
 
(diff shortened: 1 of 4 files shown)


src/pages/_includes/ui/select.html
```diff
@@ -56,7 +65,7 @@
 		var el;
 		window.TomSelect && ({% if jekyll.environment == 'development' %}window.tabler_select["select-{{ id }}"] = {% endif %}new TomSelect(el = document.getElementById('select-{{ id }}'), {
 			copyClassesToDropdown: false,
-			dropdownClass: 'dropdown-menu',
+			dropdownClass: 'dropdown-menu ts-dropdown',
```
kevinpapst, line 68: by adding this class, [the original tomselect rules](https://github.com/orchidjs/tom-select/blob/master/src/scss/tom-select.bootstrap5.scss#L76) now will be applied

rjd22 (approved):
LGTM

kevinpapst:
Ha, I already forgot about this PR 😁 

Thanks @rjd22 for merging 👍

Issue #1082, opened 2022-03-09
[BUG] bg-*color* overwrites font-color


JustGreg:
**Describe the bug**

In [utils/_colors.scss](https://github.com/tabler/tabler/blob/main/src/scss/utils/_colors.scss) font-color is set to `#ffffff !important`.
Generated CSS would look like this:
```css
.bg-dark {
  color: #ffffff !important;
  background: #1e293b !important;
}
.text-dark {
  color: #1e293b !important;
}
.bg-white {
  color: #ffffff !important;
  background: #ffffff !important;
}
.text-white {
  color: #ffffff !important;
}
```

The following HTML-Code will display a white text on a white background:
```html
<div class="bg-white text-dark">text</div>
```
(this is because bg-white is declared after text-dark)

For some reason this part is missing in https://preview.tabler.io/dist/css/tabler.css

rjd22:
@JustGreg I could not replicate your issue. I did make a fix on it based on your feedback but @codecalm needs to check if my fix is appropriate:

https://github.com/tabler/tabler/pull/1086

ishita-chandra:
[PR-1112](https://github.com/tabler/tabler/pull/1112)
I've added a pull request to fix the text colour being overwitten by bgcolour issue. Please have a look.

PR #1112, opened 2022-04-24
BUG: text colour overwritten by bg colour FIXED

ishita-chandra:
Removed the important tag so that text colour is given priority. 

[ISSUE 1082](https://github.com/tabler/tabler/issues/1082)  issue fixed


kevinpapst:
Removing `!important` is always a good idea, but now there are 2 proposed fixes. Did you try #1086 ?

rjd22:
I'm not sure removing the important is wise in this case. bg-* is used in many other places that might break with it removed.

PR #1114, opened 2022-04-29
Datatables

codecalm:


kevinpapst:
Nice that you started working on that. I already have a custom data-table implementation. What do you think aboubt an indicator class, which can be used to show that a column is sortable. 

Bildschirmfoto 2022-04-29 um 13 09 11

codecalm:
@kevinpapst done :)

kevinpapst:
Nice! I had a quick look and missed it earlier.

Can you have a look at the header in dark mode. 
The text is hidden on-hover and the icons are practically invisible.

PR #1131, opened 2022-06-06
update Bootstrap to `5.2.0-beta1`

codecalm:


kevinpapst:
Not sure if this is related to the PR, but it was working just a couple of days ago and this here was the biggest change lately. So I assume this update broke several things, one of it was modals:
https://preview.tabler.io/docs/modals.html

Unreadable in both dark and normal mode now, e.g.
Bildschirmfoto 2022-06-07 um 21 12 56

I can look for more quirks if you like. You know: I saw the PR when you opened it and was about to ask for a new release, but you were even quicker with the merge 😁 

Here is another glitch:
Bildschirmfoto 2022-06-07 um 21 21 26

I'd like to offer my help for the future. I don't mind if you ping me directly and call out for help with testing. I can only speak for myself, but I am available and happy to support in that way.

Or is this just a misunderstanding on my end and https://preview.tabler.io simply outdated?

PR #1139, opened 2022-06-08
Dropzone

codecalm:
screencapture-localhost-3001-dropzone-html-2022-06-09-00_28_51

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,6 +31,7 @@
         "countup.js": "^2.1.0",
         "critical": "^4.0.1",
         "cross-spawn": "^7.0.3",
+        "dropzone": "^6.0.0-beta.2",
         "flatpickr": "^4.6.13",
         "gulp": "^4.0.2",
         "gulp-clean": "^0.4.0",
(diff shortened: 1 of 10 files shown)


kevinpapst:
Is it supposed to work already? In Safari I cannot drop anything (cursor doesn't change as well) and it still shows the file chooser:
Bildschirmfoto 2022-06-09 um 09 05 33

codecalm:
it was a problem with `package.json`, try now :)

kevinpapst:
Nice integration, fits well and adding ten files is displayed correct 👍
Bildschirmfoto 2022-06-09 um 14 17 28

I guess the entire handling / styling is part of the library, correct?! The error is shown because there is no active counterpart on the server?

In my opinion the `X` icon is too large (it blocks the onhover infos) and doesn't work. 
Haven't checked why and not entirely necessary, but would be neat for the demo if it would remove the file. 
The "bigger issue" (first world problem 😁) for me is that the preview is not visible. 

Another idea: if the lib allows to customize allowed file/mime types, then configure "Custom drop zone" to allow jpg/png only as example.

gladtoseeugo:
UNSUBSCRIBE

PR #1143, opened 2022-06-08
Tinymce

codecalm:
screencapture-localhost-3001-tinymce-html-2022-06-09-01_00_47


sanamhub:
Thanks @codecalm for this!
I was trying to integrate TinyMCE myself, just noticed you already did!

Issue #1149, opened 2022-06-18
[BUG] Nav overflow on Vertical Layout


copulatrix:
**Describe the bug**
With the Vertical layout. The nav is overflowing.

**Device**
 - Browser:
   -  Chrome: v102.0.5005.115
   - Edge: 102.0.1245.41
   - Firefox: 101.0.1
 - OS: Windows 10 Pro N, v21H2, Build 19044.1766
 - Screen size: 1920x1080

**To reproduce**
***Light Mode***
Steps to reproduce the behaviour:
1. Go to 'https://preview.tabler.io/layout-vertical.html?theme=light'
2. While the x-axis scrollbar isn't visible, you're able to scroll left and right on the nav.

***Dark Mode***
Steps to reproduce the behaviour:
1. Go to 'https://preview.tabler.io/layout-vertical.html?theme=dark'
2. You can see scrollbar, and of course you're able to scroll left and right on the nav.

**Screenshots**
***Light Mode***
![image](https://user-images.githubusercontent.com/76273853/174430388-9fb101cc-0b09-4226-8c02-933c017d582b.png)

_Scrolled Right_
![image](https://user-images.githubusercontent.com/76273853/174430409-8a76508b-900c-4cf0-9502-f8b143807541.png)

***Dark Mode***
![image](https://user-images.githubusercontent.com/76273853/174430347-6a0e8447-1156-4ebe-8b3a-c62a4b81968f.png)

_Scrolled Right_
![image](https://user-images.githubusercontent.com/76273853/174430336-8fddef2a-19d4-4c73-94f3-1669fc573d90.png)

PR #1151, opened 2022-06-18
Fixes Issue #1149. Vertical nav overflow.

copulatrix:
Before
image

After
image


rjd22:
@codecalm could you check what the proper fix would be this one or #1155 ?

PR #1152, opened 2022-06-18
Fix ApexChart heatmap initialization on the charts docs page

dheineman:
Changed the javascript chart initialization to conform with `chart.html` and `chart-sparkline.html`


rjd22:
Looks good to me. Thanks!

PR #1154, opened 2022-06-18
Make the datatables preview page functional

dheineman:
A number of fixes for the datatables preview page (see https://preview.tabler.io/datatables.html)

- Add list.js to the devDependencies to actually make sorting columns work
- Wrap the table in `table-responsive` to show properly on tablet and mobile
  - make the progress text be full with on tablet/mobile to prevent breaking out of the table.   
- (other package.json changes are just alphabetising)


rjd22:
@dheineman Thank you!

PR #1161, opened 2022-07-02
Required form label for horizontal forms

kevinpapst:
According to the Bootstrap docs labels do get `col-form-label` in horizontal forms and not `form-label` as in classical form layout. See examples at https://getbootstrap.com/docs/5.2/forms/layout/#horizontal-form . Some frameworks that render forms (e.g. PHP Symfony) do stick to the original definition and I was missing the * in some forms.

Tabler until now does only style `form-label`, there is no mention of `col-form-label`.
This was okay in most cases, but the `required` class isn't applied (compare before and after screenshots).

I adjusted the demo, so there is a showcase of the `required` class as well in horizontal forms. I also adjusted the used label class, so the demo HTML matches the HTML from the Bootstrap docs. 

Before:
Bildschirmfoto 2022-07-02 um 23 50 17

After:
Bildschirmfoto 2022-07-02 um 23 50 41

The screenshots were taken after adjusting the HTML (see required * on the left) but before adjusting the CSS.

### DIFF
--- a/src/pages/_includes/cards/form/layout.html
+++ b/src/pages/_includes/cards/form/layout.html
@@ -6,7 +6,7 @@
 	<div class="card-body">
 		<form>
 			<div class="form-group mb-3 {% if horizontal %}row{% endif %}">
-				<label class="form-label{% if horizontal %} col-3 col-form-label{% endif %}">Email address</label>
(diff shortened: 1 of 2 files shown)


rjd22:
Thank your for fixing this!

kevinpapst:
Thanks for review and merge 👍

Issue #1147, opened 2022-06-13
Beta10 not in npm packages?


yangfancn:
I see the latest version in npmjs.com/package/ is 1.0.0-beta9

kevinpapst:
Here as well, `yarn upgrade` still installs beta9:
<img width="286" alt="Bildschirmfoto 2022-06-24 um 18 56 46" src="https://user-images.githubusercontent.com/533162/175607583-dfe8ea01-75dd-45a0-a498-ab3afd5b836b.png">

@codecalm can you push the beta-10 update to npm, so we can start testing?

kevinpapst:
For everyone else who has hard times (like me) figuring this one out, the version is indeed available but not tagged as latest.

The website shows beta9 as latest, but if you switch to the Versions page you see beta10:
https://www.npmjs.com/package/@tabler/core?activeTab=versions

So you have to set the version in your update command:
```
npm i @tabler/core@1.0.0-beta10
```

PR #1165, opened 2022-07-15
docs: switch unpkg to jsdelivr

ayuhito:
Continuing on from https://github.com/tabler/tabler-icons/pull/265.

I also noticed that the latest tag on NPM is actually linked to beta9 instead of 10 as seen here: https://www.npmjs.com/package/@tabler/core?activeTab=versions

That has a side-effect of the CDNs linking to beta9 instead of 10 when you use the `@latest` tag on it. Is that intentional?

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,13 +130,13 @@ All files included in `@tabler/core` npm package are available over a CDN.
 #### Javascript
 
 ```html
-<script src="https://unpkg.com/@tabler/core@latest/dist/js/tabler.min.js"></script>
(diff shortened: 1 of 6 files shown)


kevinpapst:
> Is that intentional?

Likely not, see here: https://github.com/tabler/tabler/issues/1147 - already reported a month ago...

ayuhito:
Yeah I don't think that's a solid solution for the long term, you can manually change the dist tag like this: `npm dist-tag add <pkg name>@<version> latest`

Otherwise, a maintainer can send a quick support ticket to NPM if they're not fully comfortable with playing around with this. The NPM support team is super helpful.

Issue #1145, opened 2022-06-09
[BUG] Multiple bugs after Bootstrap update


kevinpapst:
Not sure if this is related to the "Bootstrap beta" PR, but it was working just a couple of days ago and this here was the biggest change lately. So I assume this update broke several things, one of it was modals:
https://preview.tabler.io/docs/modals.html

Unreadable in both dark and normal mode now, e.g.
<img width="883" alt="Bildschirmfoto 2022-06-07 um 21 12 56" src="https://user-images.githubusercontent.com/533162/172463535-c6c5d068-70f2-44ba-a4ec-3e89009ac710.png">

I can look for more quirks if you like. You know: I saw the PR when you opened it and was about to ask for a new release, but you were even quicker with the merge 😁 

Here is another glitch:
<img width="545" alt="Bildschirmfoto 2022-06-07 um 21 21 26" src="https://user-images.githubusercontent.com/533162/172464832-c08329e3-20e3-429f-88db-fb8a1cc83948.png">

I'd like to offer my help for the future. I don't mind if you ping me directly and call out for help with testing. I can only speak for myself, but I am available and happy to support in that way.

Or is this just a misunderstanding on my end and https://preview.tabler.io simply outdated?

_Originally posted by @kevinpapst in https://github.com/tabler/tabler/issues/1131#issuecomment-1149081320_

dheineman:
@kevinpapst I cannot seem to reproduce the second glitch where the button text is not visible. Are you still seeing this and if so, do you have a specific way to reproduce it?

kevinpapst:
Interesting, now that you answered I tried in other browsers (Firefox works) and maybe it is a Safari only bug.

I made this screenshot right now:
<img width="847" alt="Bildschirmfoto 2022-07-30 um 13 02 20" src="https://user-images.githubusercontent.com/533162/181908014-f798a10c-4afd-484e-9686-82cbe979c0be.png">

So nothing special, just open the page.

kevinpapst:
Firefox says this:
<img width="546" alt="Bildschirmfoto 2022-07-30 um 13 05 02" src="https://user-images.githubusercontent.com/533162/181908080-6b66707c-290e-4603-9f10-03e1e1e58bf9.png">
Maybe the variable `--tblr-color-text` is missing and different browsers handle that problem in different ways?

dheineman:
> Maybe the variable `--tblr-color-text` is missing and different browsers handle that problem in different ways?

[...]

PR #1171, opened 2022-07-29
Fix rendering of modal examples on docs page

dheineman:
fix for modal bugs reported in https://github.com/tabler/tabler/issues/1145

Bootstrap 5.2 moved to using CSS variables for modals (https://github.com/twbs/bootstrap/pull/36060) which are set on the `modal` class. As this class is not included in the sample code, the other modal classes (like `modal-dialog` and `modal-content`) could not find the correct variables to set things like the padding/margin etc.

This PR will add the `modal` class to the sample code as it is required for both the end user and correctly rendering the modals in the documentation.

### DIFF
--- a/src/pages/_docs/modals.md
+++ b/src/pages/_docs/modals.md
@@ -7,9 +7,11 @@ menu: docs.components.modals
 ## Default markup
 
 {% capture code %}
-<div class="modal-dialog" role="document">
-    <div class="modal-content">
-        {% include parts/modals/simple.html %}
+<div class="modal{% hide %} d-block position-relative{% endhide %}" tabindex="-1">
+    <div class="modal-dialog" role="document">
+        <div class="modal-content">
+            {% include parts/modals/simple.html %}
+        </div>
     </div>
 </div>
 {% endcapture %}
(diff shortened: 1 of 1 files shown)


src/pages/_docs/modals.md
```diff
@@ -7,9 +7,11 @@ menu: docs.components.modals
 ## Default markup
 
 {% capture code %}
-<div class="modal-dialog" role="document">
-    <div class="modal-content">
-        {% include parts/modals/simple.html %}
+<div class="modal{% hide %} d-block position-relative{% endhide %}" tabindex="-1">
```
kevinpapst, line 10: never seen `hide`, what is it? same as `comment`?

dheineman: It is a custom tag for the documentation generation system to hide text or code in the example/demo code section whilst still using it to render the example itself. see: https://github.com/tabler/tabler/blob/main/src/pages/_includes/example.html#L2

Those two classes are only required to render the modals on the docs page as `.modal` has a `display: hidden` and `position: fixed`, which is expected behaviour as modals are hidden by default. But as they should not be added by end users we hide them in the demo code.

kevinpapst: interesting approach with the custom tag ... no wonder I couldn't find it in the liquid documentation.
thanks for the link, haven't seen that part yet 👍 

and thanks for fixing the modal, very nice!

rjd22:
Thank you for the fix!

Issue #1167, opened 2022-07-21
[BUG] The pricing table in the demo is not responsive


artknight:
The pricing table in the demo is not responsive.
https://preview.tabler.io/pricing-table.html

**To reproduce**

1. Go to https://preview.tabler.io/pricing-table.html on any mobile device

**Screenshots**

<img width="417" alt="Screen Shot 2022-07-21 at 5 12 24 PM" src="https://user-images.githubusercontent.com/157762/180316071-73f2ff0c-d973-47e1-b363-782f21a07454.png">

PR #1174, opened 2022-07-30
Fix the responsiveness of the pricing table page

dheineman:
fixes https://github.com/tabler/tabler/issues/1167

Whitespace changes seem to have been automatically applied by my IDE based on the repositories .editorconfig. Not sure if this is desired but i can revert those if not.


rjd22:
That is way better! Thank you!

PR #1179, opened 2022-08-01
Updating spelling

eryan411:
Visualise is the British spelling of visualize and inconsistent with the rest of the file (uses color instead of colour, etc), updated to match document


gladtoseeugo:
UNSCRIBE

PR #1189, opened 2022-08-05
Inter iconfont

codecalm:


kevinpapst:
There was a reason why we removed it: https://github.com/tabler/tabler/pull/1048

Can you make this a variable that can be overwritten by implementations? I am very happy with system default fonts. Wouldn't it cause the same bugs (e.g. [here](https://github.com/tabler/tabler/issues/1028)) as before, if someone has Inter installed?

codecalm:
I've removed Inter font from CSS and override it in HTML. Now everything should be fine :)

kevinpapst:
That looks like a great solution and example who to integrate custom fonts 👍

kevinpapst:
And I do not see the capital `A` issue anymore:
Bildschirmfoto 2022-08-05 um 12 19 54

PR #1190, opened 2022-08-05
Remove Google Analytics from preview

codecalm:


kevinpapst:
THANK YOU SO MUCH ❤️ Open source should IMO promote some other values as well and this PR is a good example 😄

And it will actually bring tabler.io down from one of the top positions (actually place 4) in my Safari data protection report to one of the middle positions:
Bildschirmfoto 2022-08-05 um 12 34 48

codecalm:
I'm working still on new tabler.io version which be released without google conections :) 

Screenshot 2022-08-05 at 13 31 26

kevinpapst:
Looks beautiful, you are doing a great job!

PR #1192, opened 2022-08-05
Added Docker support

vitteloil:
## Running with Docker

If you don't want to install node/npm/ruby and the dependencies on your local environment, you can use the provided Dockerfile to build a docker image.
This Dockerfile is provided as an example to spin-up a container running Tabler.

Example of how to use this image:

1. Build the tabler image : `docker build -t tabler .`
2. Run the tabler image while mounting the `src` directory as well as the `_config.yml` file into the container.

Don't forget to expose the port 3000 so you can browse the website locally.
You can also expose the port 3001 to have access to BrowserSync

```sh
docker run -p 3000:3000 -p 3001:3001 -v $(pwd)/src:/app/src -v $(pwd)/_config.yml:/app/_config.yml tabler
```

Now open your browser to [http://localhost:3000](http://localhost:3000). Edit anything in the `src/` folder and watch your browser refresh the page after it has been rebuilt.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM ubuntu:20.04
+
+# check https://github.com/nodesource/distributions for setup_XX versions
+ARG NODE_SETUP=setup_14
+ARG RUBY_VERSION=2.7
+
+# RUBY installation
+RUN apt-get update \
+    && apt-get -y install curl gcc g++ make software-properties-common \
+    && apt-add-repository ppa:brightbox/ruby-ng && apt-get update \
+    && apt-get -y --no-install-recommends install ruby$RUBY_VERSION ruby$RUBY_VERSION-dev ruby-switch \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM ubuntu:20.04
```
rjd22, line 1: Doing it this way would cause a really big docker container to be build. I would recommend using the ruby container as a base:

https://hub.docker.com/_/ruby

For example `2.7-alpine` would be a good base version that is really small.

rjd22 (changes requested):

gladtoseeugo:
UNSCRIBE

rjd22:
The comment have been fixed and merged in https://github.com/tabler/tabler/pull/1211. Thank you for the work you have done.

Issue #1193, opened 2022-08-08
[BUG] Table colors in dark mode look too bright 


dheineman:
**Describe the bug**

The table (row) colors do not look very good/integrate properly in dark mode.

**Device**

 - Browser: Chrome Version 103.0.5060.134 (Official Build) (arm64)
 - OS: macOS 12.4 (21F79)
 - Screen size: not applicable

**To reproduce**

1. Add a table as defined in the [bootstrap docs](https://getbootstrap.com/docs/5.2/content/tables/#variants) to your page
2. Enable dark mode

**Screenshots**

![Screenshot 2022-08-08 at 09 27 54](https://user-images.githubusercontent.com/516028/183363648-689a7ac4-4523-4a65-b383-c4ed336e9644.png)

rjd22:
@dheineman would you be willing to make a fix for this?

gladtoseeugo:
UNSCRIBE

gladtoseeugo:
UNSCRIBE 



From: "Dave Heineman" ***@***.***> 
To: "tabler" ***@***.***> 
Cc: "Subscribed" ***@***.***> 
Sent: Sunday, August 7, 2022 11:37:24 PM 
Subject: [tabler/tabler] [BUG] Table colors in dark mode look too bright (Issue #1193) 





Describe the bug 

The table (row) colors do not look very good/integrate properly in dark mode. 

Device 

    * Browser: Chrome Version 103.0.5060.134 (Official Build) (arm64) 
    * OS: macOS 12.4 (21F79) 
    * Screen size: not applicable 


To reproduce 

Steps to reproduce the behavior: 

    1. Add a table as defined in the [ https://getbootstrap.com/docs/5.2/content/tables/#variants | bootstrap docs ] to your page 
    2. Enable dark mode 


Screenshots 

[ https://user-images.githubusercontent.com/516028/183363648-689a7ac4-4523-4a65-b383-c4ed336e9644.png ]

dheineman:
I am not that good with SCSS but i gave it an attempt in https://github.com/tabler/tabler/pull/1200

PR #1200, opened 2022-08-11
Use dark table variant colors in dark mode

dheineman:
I am not exactly an expert on scss, variables or mixins so i am not sure if this is the desired approach but the PR will shift the colors for variant tables to shade instead of tint when in dark mode.

fixes: https://github.com/tabler/tabler/issues/1193

Screenshot 2022-08-11 at 11 32 37


rjd22:
Thank you @dheineman looks great!

Issue #1202, opened 2022-08-12
[BUG] Active item in pagination disappears on hover


kevinpapst:
**Describe the bug**

When hovering the active link in a pagination, it will be hidden due to the transparent background.

Don't know when this was introduced, but I am sure it worked in the past.
The [SCSS file wasn't touched in a long while](https://github.com/tabler/tabler/blob/dev/src/scss/ui/_pagination.scss#L10), so it might be a regression coming from another change.

**Device**

Happens in Safari and Firefox.

**To reproduce**

Hover the active pagination item at the bottom of the page:
https://preview.tabler.io/tables.html?theme=light

**Screenshots**
Regular:
<img width="305" alt="Bildschirmfoto 2022-08-12 um 14 35 35" src="https://user-images.githubusercontent.com/533162/184355105-35715cb6-a425-4907-b04c-2ad409ffa1c3.png">

On hover (mouse cursor disappears on screenshot):
<img width="1395" alt="Bildschirmfoto 2022-08-12 um 14 36 24" src="https://user-images.githubusercontent.com/533162/184355241-0d8aa420-a559-4d3d-b74e-357880e9a2b1.png">

PR #1203, opened 2022-08-12
fix transparent hover background in pagination

dheineman:
fixes: https://github.com/tabler/tabler/issues/1202


kevinpapst:
Damn, that was quick 😁 thanks ❤️  

Works for me as expected.

PR #1211, opened 2022-08-19
Added Dockerfile with alpine base

baa-ableton:
This is simply a re-work of [PR-1192](https://github.com/tabler/tabler/pull/1192).

Alpine base image was requested, and I added it. Branch works locally with MacOS, Docker Desktop with the image holding 1.12GB of space.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+demo/
+.git/
+.github/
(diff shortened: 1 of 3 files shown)


rjd22:
@baa-ableton thank you for the work! This is great!

Issue #1177, opened 2022-07-30
[BUG] Button text and icon is invisible due to missing CSS variable


kevinpapst:
**Describe the bug**

This bug seems to have different visible consequences in different browsers:
<img width="545" alt="Bildschirmfoto 2022-06-07 um 21 21 26" src="https://user-images.githubusercontent.com/533162/172464832-c08329e3-20e3-429f-88db-fb8a1cc83948.png">

Firefox says this:
<img width="546" alt="Bildschirmfoto 2022-07-30 um 13 05 02" src="https://user-images.githubusercontent.com/533162/181908080-6b66707c-290e-4603-9f10-03e1e1e58bf9.png">

These are btn-white in Firefox:
<img width="938" alt="Bildschirmfoto 2022-08-30 um 13 07 36" src="https://user-images.githubusercontent.com/533162/187421518-eb89c9f2-71cb-462d-8816-9c2588c37207.png">

I believe that the variable `--tblr-color-text` is missing and different browsers handle that problem in different ways.
The buttons icons/text are invisible in Safari, while some are visible in Firefox and Chrome.

We discussed that [here](https://github.com/tabler/tabler/issues/1145#issuecomment-1200137114) but I wanted to raise a new bug, as the other one was too generic.

**Device**

 - Safari 15.5
 - Firefox 103 & 104

**To reproduce**

Simply open https://preview.tabler.io/ in Safari

kevinpapst:
@codecalm this is a blocker for me (and likely every other Safari user). I tested Safari on Mac and iPad with latest versions, and both cannot be used in dark mode because the buttons are unreadable. I can try to fix it, but I need some ideas where to start. Any hint that you want to share?

codecalm:
fixed

kevinpapst:
thanks!

PR #1214, opened 2022-08-23
move SCSS variables to CSS variables

codecalm:
screencapture-localhost-3000-index-html-2022-08-31-21_28_15


kevinpapst:
Perfect time to look into #1177 😁

Issue #1219, opened 2022-08-30
[BUG] .btn-white - theme-dark


wangkanai:
**Describe the bug**

`.btn-white` does not invert color when in `.theme-dark`

**Device**

 - Browser [Google Chrome v104]
 - OS: [Windows 10 Pro]
 - Screen size [e.g. 3840 x 2400]

**To reproduce**

1. Go to [Tabler preview -> Form element ](https://preview.tabler.io/form-elements.html?theme=dark)
2. Click on 
![image](https://user-images.githubusercontent.com/10666633/187372722-4168655a-fdec-45e1-b2f3-96cc6ad286df.png)

3. Scroll down to 'Toolbar'

![image](https://user-images.githubusercontent.com/10666633/187372148-b19bbdff-b52e-43ff-a881-c77bfa3260d6.png)

If applicable, add screenshots to help explain this problem.

PR #1220, opened 2022-08-30
fixed btn-white dark mode theme

wangkanai:
Reference: #1219 [BUG] .btn-white - theme-dark

Create an invert color for `btn-white` with in dark mode.

image

Wow, I like this vercel preview feature


kevinpapst:
This button suffers the same issue as other ones in dark mode. 
See https://github.com/tabler/tabler/issues/1177
The underlying issue should be fixed first.

Just as a side note: should white buttons really turn into regular colored buttons?

wangkanai:
bootstrap v5.2 has change the ::root variable to `--#{$prefix}body-color: #{$body-color};` for text color. Changing this would fix issue #1177 also, 

image

In which you can now resolve this issue.

image

This is from investigating the root cause.

image

Therefore, after all the corrections we would have much more clean css stylesheet.

image

rjd22:
@wangkanai @kevinpapst `.btn-white` buttons are a bit weird. Normal buttons with that class seem to change in dark mode so I would guess fixing this for the other buttons is acceptable.

But in my personal opinion `.btn-white` buttons should not change but since this is already the case I'm okay with merging this.

rjd22:
@wangkanai a lot of other changes came with your commits. Could you please make sure only the needed changes are in this PR?

kevinpapst:
@wangkanai I am just an interested user of the theme, but both my frontend skills and background knowledge about the maintainers ideas for this theme are very limited. If you really fixed the btn issue, that would be awesome!!!!!!

@rjd22 If you check the button documentation all colored buttons keep their color in dark mode. Unfortunately nothing is stated about their behavior in dark mode. On "form elements" page the btn-white buttons stay white. My impression was that these explicit colored buttons keep their color in both (dark/light) modes.
But that is obviously just my personal interpretation.

And yes I agree, that btn-white looks creepy in dark mode ^^ I had to add a workaround in my project for a white button in overlapping navigation, which looks like shit in dark mode. So I'd like to see the change myself, just still questioning if it is good to introduce that BC. Isn't there some color class that could be used like `btn bg-white-in-light-only` for that feature?

rjd22:
@kevinpapst For example the button with the text "New View" on the top right is dark colored in dark mode but contains `.btn-white` on https://preview.tabler.io/ .

This is the reason I think @codecalm intention is that `.btn-white` color buttons are dark in dark mode even if they are then really weirdly named.

rjd22:
@wangkanai there are still a lot of unrelated changes in there. Maybe it might be easier to open a new branch from `dev` and reapply the patches you made?

PR #1228, opened 2022-08-31
Fixed dark mode theme .btn from `color-text` to `body-color`

wangkanai:
This PR is recreate to #1220 fixed btn-white dark mode theme

bootstrap v5.2 has change the ::root variable to `--#{$prefix}body-color: #{$body-color};` for text color. Changing this would fix issue #1177 also, 

image

In which you can now resolve this issue.

image

This is from investigating the root cause.

image

Therefore, after all the corrections we would have much more clean css stylesheet.

image

What do you think @kevinpapst ?

### DIFF
--- a/src/scss/ui/_buttons.scss
+++ b/src/scss/ui/_buttons.scss
@@ -65,8 +65,8 @@
 // Button color variations
 //
 .btn-white {
-  --#{$prefix}btn-bg: var(--#{$prefix}card-bg);
-  --#{$prefix}btn-hover-bg: var(--#{$prefix}card-bg);
+  //--#{$prefix}btn-bg: var(--#{$prefix}card-bg);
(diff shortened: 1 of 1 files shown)


rjd22:
Thank you for the time you're offering to fix this. I will take a deeper look and test this tomorrow.

wangkanai:
@rjd22 Your welcome, I really like you guys have done here. I have also made Blazor template and components that based on Tabler natively. [Wangkanai.Tabler](https://github.com/wangkanai/wangkanai/tree/main/Tabler) while is still in a working progress.

codecalm:
hi @wangkanai! thank you for your contribute

everything looks good, but default buttons has transparent background now, should have white (or color like `.card-bg`)
 
Screenshot 2022-08-31 at 22 31 40

wangkanai:
I'm fixing the whole issue with PR #1229 Fixed .card global variable --#{$prefix}card-bg

PR #1250, opened 2022-09-15
Changed Dropdown Min-Width to 13rem

NageshMandal:


kevinpapst:
You should post your SCSS changes and NOT the compiled/minified code.
Any experienced maintainer should/would reject a PR, as the min file could potentially contain any kind of new content.

PR #1264, opened 2022-09-20
Switch to dart-sass

ihmels:
In #1213 the Sass implementation was switched from dart-sass to node-sass for some reason. Because node-sass is deprecated we should switch back to dart-sass. Also, it should only be listed in the devDependencies.


codecalm:
oh, I've tested sass and accidencially revert `sass` to `node-sass`. Thank you! ❤️

PR #1266, opened 2022-09-20
update Bootstrap to 5.2.1, update dependencies

codecalm:


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0-beta12",
   "description": "Premium and Open Source dashboard template with responsive and high quality UI.",
   "homepage": "https://tabler.io",
+  "type": "commonjs",
```
ihmels, line 6: I think this line should be removed (this source should use `import` instead of `require()`.

ihmels:
Can't we remove the critical library? The benefit doesn't seem very high to me for the demo and this version of the library pulls in old dependencies with security holes.

codecalm:
> Can't we remove the critical library? The benefit doesn't seem very high to me for the demo and this version of the library pulls in old dependencies with security holes. 🤔

yes, you're right, I've removed it

Issue #1249, opened 2022-09-15
[BUG] Dropdown issue on tables


adolfoflores:
**Describe the bug**

Dropdown is not working correctly in tables.

**Device**

Tested in Chrome and Safari.

**To reproduce**

You can reproduce it in the demo page: https://preview.tabler.io

<img width="570" alt="CleanShot 2022-09-14 at 22 05 07@2x" src="https://user-images.githubusercontent.com/2294172/190304391-2bf19c97-dc8c-4b59-ac58-b5a07221e184.png">

NageshMandal:
I want to work on this issue. Could you please assign it to me.

adolfoflores:
> I want to work on this issue. Could you please assign it to me.

I think you can fix it and make a pull request.

NageshMandal:
> > I want to work on this issue. Could you please assign it to me.
> 
> I think you can fix it and make a pull request.

Could you please explane the Dropdown issue In More Details

adolfoflores:
> > > I want to work on this issue. Could you please assign it to me.
> > 
> > 
> > I think you can fix it and make a pull request.
> 
> Could you please explane the Dropdown issue In More Details

1. Go here: https://preview.tabler.io
2. Scroll down until the "Invoices" table.
3. Click on "Actions" dropdown.

You'll see how is not displaying correctly. Looks like a z-index issue, but is not.

<img width="235" alt="CleanShot 2022-09-14 at 22 54 42@2x" src="https://user-images.githubusercontent.com/2294172/190310246-f020a931-1148-45dc-81ab-2d06a1228688.png">

fahad-islam:
@adolfoflores Do you want it like this one?

![Screenshot from 2022-09-15 08-41-46](https://user-images.githubusercontent.com/89579421/190316534-4abde5ee-f5b5-41af-a528-bd47423cd157.png)

Santanu-dev:
This change needs to be done using boundary options
https://popper.js.org/docs/v2/utils/detect-overflow/#boundary

PR #1268, opened 2022-09-20
fixed dropdown issue on tables

Santanu-dev:
#1249 
afterChange

### DIFF
--- a/src/scss/ui/_dropdowns.scss
+++ b/src/scss/ui/_dropdowns.scss
@@ -1,7 +1,7 @@
 .dropdown-menu {
   box-shadow: $box-shadow;
   user-select: none;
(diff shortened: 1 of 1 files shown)


ihmels:
The problem should not be solved in this way. It must be solved via the [boundary option][1].

[1]: https://getbootstrap.com/docs/5.2/components/dropdowns/#options

Santanu-dev:
> The problem should not be solved in this way. It must be solved via the [boundary option](https://getbootstrap.com/docs/5.2/components/dropdowns/#options).

[boundary option] is required for only this table, right? 
and for all other dropdowns it should be default one.

nalin:
I can't find a boundary option that prevents the clipping. Any suggestions?

PR #1288, opened 2022-09-28
Profile page

codecalm:


src/pages/_layouts/default.html
```diff
@@ -19,7 +19,7 @@
 		{% include layout/page-header.html %}
 
 		<!-- Page body -->
-		<div class="page-body">
+		<main class="page-body{% if page.page-body-class %} {{ page.page-body-class }}{% endif %}">
```
kevinpapst, line 22: May I ask: is there a strong reason for that change? It seems like a BC break, especially regarding custom CSS rules.

codecalm: it's temporary

kevinpapst:
I like the new page. Simple and effective. I guess by adding a second card on the right we could add links to other profile sections. Or by using the top level buttons... Oh and I sincerely hope that Dunn Slane's intro text has some truth for your projects in it 😄 

One wish, which I want to explain: 
The new image cover does not fit perfectly into the overall look for tabler (just my very personal opinion). I would like to remove (if I would use the page) and as first approach tried to remove the `page-cover` node, but the the profile image moved upwards to a completely wrong position. 

And now my wish:
Any chance it could be build in a way where the cover pushes the content down, while keeping the image at the same position, so we could use the page very easily with and without the cover?

src/pages/profile.html
```diff
@@ -0,0 +1,69 @@
...
+				</div>
+
+				<ul class="timeline mt-5">
+					{% for item in site.data.timeline %}
+					<li class="timeline-event">
+						<div class="timeline-event-icon bg-{{ item.icon-bg }}">
+							{% include ui/icon.html icon=item.icon %}
+						</div>
+						<div class="timeline-event-card card">
+							<div class="card-header card-header-light">
```
kevinpapst, line 52: `card-header-light` seems to behave wrong in dark mode

kevinpapst:
Is there a class that turns the bright white icons in dark mode a little darker?
Bildschirmfoto 2022-11-14 um 23 53 20

Everything else is 👍

codecalm:
fixed! :) 

Screenshot 2022-11-15 at 00 25 31

src/pages/_layouts/default.html
kevinpapst: I never came across the `main` element, but adapting to the standards is not a bad idea. but: this is 100% sure a BC break. Even in my own projects I found a `document.querySelector('div.page-body')` and I am only one of now 32k stargazers^^ 

So: revert `main` vs `div` or keep it?

codecalm: I reverted it, it was only temporary

kevinpapst (approved):
LGTM 🚀

Issue #1273, opened 2022-09-22
[BUG] datepicker background color broken.


adolfoflores:
**Describe the bug**

The background color of the date picker is broken.

Broken in Chrome (Mac)
Working good in Safari (Mac)

**To reproduce**
Enter to: https://preview.tabler.io/form-elements.html

**Screenshots**
<img width="589" alt="CleanShot 2022-09-21 at 23 20 01@2x" src="https://user-images.githubusercontent.com/2294172/191657783-6c970111-645d-4788-83b0-acd91b01046d.png">

malyounes:
Also broken on Chrome (Windows). 
I'm going to take a stab at it and put up a PR.

codecalm:
fixed :)

PR #1289, opened 2022-10-02
Fix for datepicker background

malyounes:
Fixes issue #1273.

Datepicker background color should now be white for light mode and respond appropriately for dark mode.


kevinpapst:
While it fixed the issue for the dropdowns, this seems to have introduced a tiny new problem in dark mode. The border corners now have a white background:
Bildschirmfoto 2022-10-03 um 10 23 40

codecalm:
I've fixed it in https://github.com/tabler/tabler/commit/592eca2beede1ebd005cbc1e09d5653e1d858583

PR #1327, opened 2022-10-23
soft-wrap the code

MatteoBira:
no changes to the code


codecalm:
this file `*.min.css` is minified by compilation. If you change it in next release will be minified again. 

If you wan unminified dile use file without `.min` extension :)

MatteoBira:
ok, thanks I didn't know it

PR #1329, opened 2022-10-24
`<kbd>` styling

codecalm:


kevinpapst:
Just a random example... 
Bildschirmfoto 2022-10-24 um 20 09 47
LGTM 👍 

But I'd add an example to the Typography page anyway, which makes it much easier to verify future changes.

Issue #1290, opened 2022-10-04
[BUG] Usage of '~' in imports is deprecated


asharnadeem:
**Describe the bug**

When importing tabler core into `styles.scss` in order to override colors, the styling breaks and warnings are thrown that the usage of tildes when importing is deprecated. Any help is appreciated. I can trudge onwards by just using the base theme colors, but would need to eventually override them with the brand colors.

**Device**
 - OS: MacOS

**To reproduce**

1. Create a new angular project
2. Add this line to `styles.scss`: `@import "@tabler/core/src/scss/tabler.scss";`
3. Run `ng serve`

**Screenshots**
![Screen Shot 2022-10-03 at 8 45 48 PM](https://user-images.githubusercontent.com/47264377/193710739-5ad0de40-bfef-42c5-96eb-eeb168d93d90.png)

nerijunior:
Are you using webpack? If so you can configure an alias to solve it. I faced the same problem when using Vite, my solution was to import all tabler files manually and all bootstrap dependencies too.

asharnadeem:
I am not using webpack. Are there plans to address this bug going forward without needing to manually import every file?

nerijunior:
@asharnadeem did you solved or imported entirely?

PR #1333, opened 2022-10-25
Prevent deprecation warning for Angular 13 or higher

lvanstrijland:
I use tabler for my Angular project and since Angular i'm getting bombed with deprecation warnings when I try to compile my project because of the tilde that is used in the scss imports.

afbeelding

So I changed the way of importing from the node_modules folder.
Please let me know what you think about this approach.

Kind Regards,
Laurens van Strijland

PR #1334, opened 2022-10-25
addresses issue #1290

copulatrix:


kevinpapst:
Two PRs for the same issue. And with different solutions. See https://github.com/tabler/tabler/pull/1333

Why do you prefer your solution?

copulatrix:
Well this is embarrassing...

In my local dev env this worked no problem. But the Vercel build didn't compile.

> Two PRs for the same issue. And with different solutions. See #1333
> 
> Why do you prefer your solution?

I wasn't aware of #1333, I don't have an interest in, nor anything to do with Angular, so I [stupidly] didn't look at it.

kevinpapst:
I wasn't saying the other PR is better, I was just curious. Seems both PRs failed to compile...

PR #1338, opened 2022-10-27
New `photogrid` component

codecalm:
screencapture-localhost-3000-photogrid-html-2022-10-28-00_22_59


kevinpapst:
Is it expected that all images in a row scale to the same height?
The last image (right side) in the first row is missing a couple of pixels and the last one in the second row seems to have these pixels too much.

codecalm:
@kevinpapst now looks better: 

screencapture-localhost-3000-photogrid-html-2022-11-12-14_24_25

I've also added lightbox to photos:

Screenshot 2022-11-12 at 14 25 20

kevinpapst:
Agree, look very nice @codecalm 

Nitpicking is annoying, I know 😁  but question again: shouldn't they align properly if the same sizes are used?

I changed the image classes and landed here:
Bildschirmfoto 2022-11-12 um 14 28 47
Bildschirmfoto 2022-11-12 um 14 31 38
Bildschirmfoto 2022-11-12 um 14 32 20

The height seems to be made from the largest image in the column set. Does the height depend on the original image size?
I would have expected that 1x1 and 2x1 have the same height of 320px and only a different width.

I don't have use-cases for such a grid, so this feedback is just from an external view point: the grid setup seems to be rather complicated, a lot of html required with a manual definition of columns/squares . You cannot just throw images at it.
I really never have to present images, so I cannot share any experience, but I have seen javascript libraries that solve the problem with some magical size adjustments and regrouping and less required html. 

But if this is simply a showcase of what Tabler is capable of: great 👍

codecalm:
it's a problem with grid gap: images with ratio (ew 2x1) need extra space for gap, which is impossible to recalculate with flex. It's first version of photogrid, I want to create more advanced component in future :) Now I used only existing classes :)

kevinpapst:
Agree, you did a good job rearranging the images in a clever order, so it is not visible ;-)

I am simply a fan of proper feedback, QA is sometimes hard if you work by yourself

PR #1349, opened 2022-11-07
Fix sidebar narrow when scrollbar appears

esl51:
When a scroll bar appears, the sidebar narrows. It can be a little annoying.
We can hide the scroll bar while retaining the behavior.

### DIFF
--- a/src/scss/layout/_navbar.scss
+++ b/src/scss/layout/_navbar.scss
@@ -388,6 +388,11 @@ Navbar vertical
             @include transition(transform $transition-time);
             overflow-x: auto;
             padding: 0;
+            scrollbar-width: none;
+
+            &::-webkit-scrollbar {
+              display: none;
+            }
 
             &.navbar-right {
               left: auto;


kevinpapst:
IMO this is bad UX. Everyone is used to scrollbars => they indicate that a part on element is currently hidden.
You loose that context if you hide the scrollbar. 

Not all users even know how to scroll down without a scrollbar. Some don't have a touchpad or mouse-wheel or just don't know how to use it. I have seen many people who are unable to get to such hidden parts of navigation-bars.

My vote goes against that: do not change common interface elements for the sake of visual appearance.

BTW there is a branch that you should check out: https://github.com/tabler/tabler/pull/1319

esl51:
> Not all users even know how to scroll down without a scrollbar. Some don't have a touchpad or mouse-wheel or just don't know how to use it. I have seen many people who are unable to get to such hidden parts of navigation-bars.

Okay, that makes sense. Then this option is not suitable. In that case, I suggest adding a left padding as wide as the scroll bar (for symmetry) and turning on the scroll bar permanently. Looks good.

```scss
padding-left: .5rem;
overflow-y: scroll;
```

> BTW there is a branch that you should check out: #1319

It's exactly the same behavior there.

kevinpapst:
> It's exactly the same behavior there.

True. I meant that you might want to target that branch, to reduce the risk of possible merge conflicts.

In my opinion "normal browser behavior" is perfectly fine and such a adjustment should be done in project code instead of patching tabler directly. But I understand that such glitches like the jumping navigation can be annoying if you concentrate too much on it - once seen you cannot unsee it 😁

codecalm:
For me it's only temporary solution. Better is redesign the sidebar so that the scrollbar does not appear.

rjd22:
I agree with @kevinpapst. We should not compromise user experience in favor for looks. Scroll bars are a important part of the UX and there are many users that are not as tech savy as we are.

Most of my users for example are artists. They often have a hard time navigating software so an predictable UX is extremely important.

PR #1353, opened 2022-11-10
fix #1348: Make job listing responsive for smaller devices

### LINKED ISSUE #1348: [BUG] Job listing is not responsive
image
Job listing is not responsive.

rjd22:
2022-11-10T14:34:22,959515063+01:00

### DIFF
--- a/src/pages/job-listing.html
+++ b/src/pages/job-listing.html
@@ -84,14 +84,14 @@ menu: extra.job-listing
 									{% if job.salary %}<div class="col-auto fs-h3 text-green">{{ job.salary }}</div>{% endif %}
 								</div>
 								<div class="row">
-									<div class="col">
+									<div class="col-md">
 										<div class="mt-3 list-inline list-inline-dots mb-0 text-muted">
 											<div class="list-inline-item">{% include ui/icon.html icon="building-community" class="icon-inline" %} {{ job.company }}</div>
 											<div class="list-inline-item">{% include ui/icon.html icon="license" class="icon-inline" %} {{ job.type }}</div>
 											<div class="list-inline-item">{% include ui/icon.html icon="map-pin" class="icon-inline" %} {{ job.location }}</div>
 										</div>
 									</div>
-									<div class="col-auto">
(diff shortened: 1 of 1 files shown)


kevinpapst:
What about `col-md-3 order-2` and `col-md-9 order-1`

The main part of this page are the listing entries, so why not moving them to the top?
Depends on the use-case obviously, but the search form on top looks "creepy" as it is now.

rjd22:
@kevinpapst having the filters on top is good UX but I agree that it is cluttered at it is now.

Normally you will hide most of the filters and limit the amount of visible filter methods to the really useful ones. In my applications I have the a show hide filters button that allow to open these more complex filters:

2022-11-10T15:21:33,168279667+01:00

kevinpapst:
Agree, I was just speaking about the state as-is. 

It is definitely better than before, but like other pages this one is far away from being optimized for small screens.

Bildschirmfoto 2022-11-10 um 15 27 51

I guess one step after another ^^

kevinpapst (approved):

rjd22:
@kevinpapst Completely agree! I do find it hard to balance where the line is for helping our users. I see Tabler more as a interface toolkit that gets you started but requires you to do your own UX.

If you find it a improvement, I could make a limited non-functional mobile version for this page with most filters hidden except the most important ones. What filters would you figure would be the most useful ones?

kevinpapst:
That is much better already! The logo could probably get full size on xs, but details...

> I do find it hard to balance where the line is for helping our users. I see Tabler more as a interface toolkit that gets you started but requires you to do your own UX.

Me too. But as framework it should showcase best practices. And the better the examples, the less googling is required for people like me^^ 

Just to give an example: I just a neat example of using tab-pills after an unnecessary long period of try & error. If such a card would have been mentioned on the tabs page, I would have easily saved 2 hours for other things ... 
I'll send an example soon for demonstration purpose.

codecalm (approved):

codecalm:
Looks good, @rjd22 can you resolve conflicts? :)

PR #1384, opened 2022-11-23
`bootstrap` import fix

codecalm:


ashgibson:
Can this commit be released soon? Currently this package can't be used on a brand new angular app due to the tilde (~) imports.

kicaj (commented):
Related #1456 

This solution doesn't work when you have not standard project structure.
Tilde (`~`) mean "start from node_modules/ in root of project"

Issue #1392, opened 2022-11-30
[BUG] Advanced Select should have a padding right


promatik:
**Describe the bug**

When there is no minimum width for an advanced select, the carret overlaps the text, like this;

![image](https://user-images.githubusercontent.com/1838187/204810655-c99fae92-2b3e-459a-9920-f5c4e5b2175a.png)

**Device**

 - Chrome 107 (but it happens everywhere)
 - Windows 10
 - 4k

**To reproduce**

Steps to reproduce the behaviour:
1. On demo, go to the forms page; https://preview.tabler.io/form-elements.html
2. Add a `width: fit-content;` to the "Select with flags" `form-select` element.

![image](https://user-images.githubusercontent.com/1838187/204811757-eb160965-5550-4753-beea-6412885774aa.png)

promatik:
Added the fix on #1393.

kevinpapst:
@promatik as both Tomselect and Tabler PRs were merged, you might wanna give it a try and close the issue if it works?!

promatik:
Hi @kevinpapst!
Both PRs were merged, none was tagged 😅
Anyway, yes it works on dev, let's assume we'll have a v1 (non beta) soon 🎉

I'll close this one for now 👌

PR #1393, opened 2022-11-30
Fix non full width selects #1392

promatik:
This is a fix for the issue reported in #1392.

With this fix, select inputs may now be `fit-content`.

image

### DIFF
--- a/src/scss/vendor/_tom-select.scss
+++ b/src/scss/vendor/_tom-select.scss
@@ -3,6 +3,11 @@ $input-border-width: 1px;
 
 @import "~tom-select/src/scss/tom-select.bootstrap5.scss";
 
+::root {
+  --ts-pr-clear-button: 0rem;
+  --ts-pr-caret: 0rem;
+}
+
 .ts-input {
   color: inherit;
 }
@@ -13,5 +18,14 @@ $input-border-width: 1px;
   .dropdown-menu {
     width: 100%;
     height: auto;
+  }
+}
+
+.ts-wrapper {
+  &.is-invalid,
+  &.is-valid {
+    .ts-control {
+      --ts-pr-clear-button: 1.5rem;
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


src/scss/vendor/_tom-select.scss
```diff
@@ -3,6 +3,11 @@ $input-border-width: 1px;
 
 @import "~tom-select/src/scss/tom-select.bootstrap5.scss";
 
+::root {
```
promatik, line 6:
```suggestion
:root {
```

promatik:
This can now be tested at; https://tabler-git-fork-promatik-patch-1-tabler-io.vercel.app/form-elements.html

With the `width: fit-content;` on the `div.form-select`.

kevinpapst:
I'd like to test that later this week with my project. I had quite some troubles with the select size in the past...

kevinpapst:
I do not use `style="width:fit-content"` anywhere for selects in my project, so I did manual tests and those were succesful.

BUT: is that only related to tabler or something that needs to be reported upstream?
Trying `width:fitcontent` on e.g. https://tom-select.js.org/examples/styling/ (select in the upper right) the issue seems to be in Tomselect.

promatik:
@kevinpapst you are right 👌
The issue is **partially** in Tom Select, I pushed this PR to fix that there; https://github.com/orchidjs/tom-select/pull/530.

Anyway, it doesn't fix everything, so this PR is still useful.
I'll add a note in the PR 🙌

src/scss/vendor/_tom-select.scss
```diff
@@ -3,6 +3,11 @@ $input-border-width: 1px;
 
 @import "~tom-select/src/scss/tom-select.bootstrap5.scss";
 
+:root {
+  --ts-pr-clear-button: 0rem;
+  --ts-pr-caret: 0rem;
+}
+
```
promatik, line 10: This will not be necessary as soon as Tom Select merges https://github.com/orchidjs/tom-select/pull/530.

PR #1399, opened 2022-12-04
Replace hard to see flag in documentation

tinyoverflow:
This PR replaces the PL flag in the Flags documentation with the rather neutral antarctica flag, as the white part of the PL flag is hard to see on the white background, which doesn't look too good in this example.

Before:
grafik

After:
grafik


codecalm:
its a problem with flags: should have a border
but your PR is good, thank you! :)

PR #1405, opened 2022-12-08
Fixed rgba conversion bug

ajaidanial:


ajaidanial:
Found the issue when converting SCSS to CSS using django-compressor & django_libsass. Let me know your thoughts.

codecalm:
thank you for your help! ❤️

PR #1419, opened 2022-12-27
fix autofill text color, not matching in dark mode

copulatrix:
Fixes Autofill text on Chrome.

Apparently due to security concerns, Chrome has limited customisation of prefilled Autofill content.

This PR simply adds `-webkit-text-fill-color: var(--#{$prefix}body-color);`, which seems to match the text color, both when typed and auto filled. 

**Before:**
image

**After:**
image


codecalm:
thank you for your help! ❤️

Issue #1410, opened 2022-12-12
too few space between card-actions


nerdoc:
**Describe the bug**

when 2 or more card-actions are displayed, there is too few space between them, compared to the other elements in Tabler.
![grafik](https://user-images.githubusercontent.com/2955584/207040434-c8c802a5-e7f1-4afc-bd0d-46d436b03bbd.png)

I helped myself with a small fix:
```css
.list-group-item-actions {
    /* add a me-1 */
    margin-right: 0.25rem !important;
}
```
Now it looks better:
![grafik](https://user-images.githubusercontent.com/2955584/207040619-50918a9d-8dba-4769-b3ee-2e2816171e2f.png)

But this should be fixed in Tabler itself... Maybe even a me-2 could be better. An I think the last one doesn't need to have this - but you know better here, @codecalm.

tinyoverflow:
Please share your HTML you use. I couldn't replicate this issue.

kevinpapst:
There are dedicated classes to style card actions, grab the required HTML from https://preview.tabler.io/card-actions.html

In short it is:
```html
<div class="card-actions btn-actions">
    <a href="#" class="btn-action">...</a>
    <a href="#" class="btn-action">...</a>
</div>
```

Looks like that:
<img width="708" alt="Bildschirmfoto 2023-05-14 um 23 57 30" src="https://github.com/tabler/tabler/assets/533162/00c5b97d-af6f-4b17-86e2-2c5a65d0084c">

PR #1420, opened 2023-01-01
Give space between card-actions

rohini-ranjanR:
Giving some space between card actions solving issue #1410


kevinpapst:
Can you please share a before and after screenshot?
Best is to have a visual somewhere in the examples.

PR #1427, opened 2023-01-07
New docs

codecalm:
Todo:
- [x] new docs engine
- [ ] better examples
- [x] search
- [ ] better documentation

DEMO: https://dev-docs.tabler.io

Preview:
Screenshot 2023-01-07 at 01 51 44


kevinpapst:
Pros: I like the look & search. Nextra has some nice features build in.

Cons: Feels weird to use something else than Tabler, like you don't trust your own framework. 

Integrating a search is possible in other ways as well, which are even compatible with GH pages and markdown/liquid. I haven't used any of those solutions yet, but there are [some](https://github.com/jekylltools/jekyll-tipue-search) if you google.

I am sorry for my downvote, but I feel that this move is not clever. Improve tabler with the missing bits instead of using an entirely different framework. To me, this feels a bit like cheating 😁 if you present the docs and users think that they can built something like it with Tabler. 

Please don't be offended, I assume that was a couple of hours work ... and I don't want to sound harsh or like an ass, but my opinion: that is not Tabler anymore and I strongly suggest that Tabler docs are built with Tabler itself. Be proud of your fantastic framework! 

If you need help building some of the Nextra features: ask. I think there are a couple of devs willing to help out.

ollym:
I agree with @kevinpapst, whatever features you found in Nextra that you liked enough to make this PR, focus on adding those features back into tabler and use tabler for your docs.

kevinpapst:
Alright, thanks for the clarification 👍 

I have some points I'd like to raise regarding the documentation, as I am managing a larger documentation project myself. Any chance we can join the discussion early?

codecalm:
yes, sure. If you have any ideas how to improve it give me know anytime :)

Issue #1456, opened 2023-02-04
[BUG] Upgrade to 17 from 16 using webpack doesn't work


kicaj:
Configuration based on webpack.

**package.js:**
```
{
  "name": "test",
  "dependencies": {
    "@tabler/core": "^1.0.0-beta17"
  },
  "devDependencies": {
    "css-loader": "^6.7.3",
    "node-sass": "^8.0.0",
    "postcss-loader": "^7.0.2",
    "sass-loader": "^13.2.0",
    "style-loader": "^3.3.1",
    "webpack": "^5.75.0",
    "webpack-cli": "^5.0.1"
  }
}
```

**webpack.config.js:**
```
const path = require('path');

module.exports = {
  entry: './index.js',
  output: {
    filename: 'main.js',
    path: path.resolve(__dirname, './dist'),
  },
  module: {
    rules: [
      {
        test: /\.(scss)$/,
        use: [
          'style-loader',
          'css-loader',
          {
            loader: 'postcss-loader',
            options: {
              postcssOptions: {
                plugins: function () {
                  return [
                    require('autoprefixer')
                  ];
                }
              },
            },
          },
          'sass-loader',
        ],
      },
    ],
  },
};
```
**index.js:**
```
import './index.scss';
import '@tabler/core';
```
**index.scss:**
```
// Variables
$assets-base: '~@tabler/core/dist';

// Imports
@import '@tabler/core/src/scss/tabler';
```

Everything works with Tabler version 1.0.0-beta15 and older.

**Docker run with:**
 - node v18.14.0
 - npm 9.3.1
 - webpack 5.75.0
 - webpack-cli 5.0.1

**To reproduce**

1. Install packages
2. Create 4 files like above
3. Run command `npx webpack`

**Problems**
```
SassError: File to import not found or unreadable: ../../node_modules/bootstrap/scss/functions.
        on line 2 of node_modules/@tabler/core/src/scss/_bootstrap-config.scss
        from line 7 of node_modules/@tabler/core/src/scss/_config.scss
        from line 1 of node_modules/@tabler/core/src/scss/_core.scss
        from line 1 of node_modules/@tabler/core/src/scss/tabler.scss
        from line 5 of index.scss
>> @import "../../node_modules/bootstrap/scss/functions";
```

**Fixes:**
Replace path to bootstrap scss files by "bootstrap/..." or "~bootstrap/..." instead of "../../bootstrap/...".

kicaj:
[...]

PR #1457, opened 2023-02-04
#1456 Fixed paths to bootstrap scss files

kicaj:


kevinpapst:
This is a duplicate of #1333 with a slightly different approach

rjd22:
@kevinpapst this feels like the correct approach with the node_modules folder excluded. This way you leave figuring out the location to your builder AFAIK?

kevinpapst:
I'd say so too, at least I have never seen a setup with `node_modules/` in the import/require statement.

rjd22:
Since the build is failing I want to make sure. @codecalm could you check if this is the right way? I get the weird underbelly feeling that it should be `@bootstrap` to make this work properly cross platform.

kevinpapst:
Isn't the @ used in case the package is scoped, which bootstrap isn't? Maybe I am totally wrong here 🤷

rjd22:
> Isn't the @ used in case the package is scoped, which bootstrap isn't? Maybe I am totally wrong here shrug

Ah, you're completely right. I thought bootstrap was scoped. But after checking it seems it isn't.

kicaj:
Everything will be better then that  `../../`

codecalm:
what if we copy bootstrap files to `scss` directory instead use it as dependency?

kicaj:
I thinks it is not good idea.

SyntaxJoker:
This problem with imports appeared in many issues and PRs, I created a separate PR to address all of them
https://github.com/tabler/tabler/pull/1477

PR #1468, opened 2023-02-12
new page: Cookie banner

codecalm:
Screenshot 2023-02-12 at 21 15 08

### DIFF
--- a/src/pages/_data/menu.yml
+++ b/src/pages/_data/menu.yml
@@ -174,6 +174,9 @@ extra:
   icon: star
   columns: 2
   children:
+    cookie-banner:
+      title: Cookie banner
+      url: cookie-banner.html
     activity:
       title: Activity
       url: activity.html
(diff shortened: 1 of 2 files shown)


rjd22:
@codecalm It would be an good idea to give an good example for this without dark patterns. So having 2 buttons with equal color priority.

The first button would be "Essential Cookies Only"
And the second button "Allow All Cookies"

The EU is quite clear on asking for permission and denying permission should be equally easy as granting permission.

codecalm:
@rjd22 I've updated it :) 

Screenshot 2023-02-21 at 15 04 32

kevinpapst:
I like the updated version 👍 Adding `btn-primary` would be good, especially in dark mode, but too late ... already merged 😁

codecalm:
Screenshot 2023-02-21 at 15 11 24

I've fixed it in dev :)

codecalm:
dark mode too: 

Screenshot 2023-02-21 at 15 12 50

Issue #1475, opened 2023-02-18
[BUG] @import "~bootstrap/scss/functions"; is broken


Dibbyo456:
**Describe the bug**

Creating new issue as requested by @tinyoverflow
Discussion: https://github.com/tabler/tabler/issues/1394#issuecomment-1434999073 

````
[plugin:vite:css] [sass] Can't find stylesheet to import.
  ╷
1 │ @import "~bootstrap/scss/functions";
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
````

**To reproduce**

Steps to reproduce the behavior: https://github.com/Dibbyo456/tabler-bug-report

SyntaxJoker:
We will be fixing that in https://github.com/tabler/tabler/pull/1477

Dibbyo456:
It works now. Thank you.

twf-nikhila:
@harryqt @kevinpapst I used npm install tabler, using the latest version
"@tabler/core": "^1.0.0-beta20",
and facing this issue

Is this not released yet?

harryqt:
No, you need to use the `dev` branch.

twf-nikhila:
@harryqt how to use dev branch when installing using npm?

harryqt:
`npm install git+https://github.com/tabler/tabler.git#dev`

twf-nikhila:
> `npm install git+https://github.com/tabler/tabler.git#dev`

It's weird...still doesn't work...using Webpack does have to do something?

PR #1466, opened 2023-02-11
Update bootstrap path in config.scss file

vinayuttam:
As part the PR #1384, the bootstrap file import path have been fixed. However, the `_config.scss` still has the old reference and this PR is address it.

Issue #1394, opened 2022-12-02
Tabler with Laravel 9 + Vite


yjungles:
Has anyone found a solution for using Tabler with Laravel 9+Vite?

For now, I'm just importing the CDN directly in the layout, overriding ::root variables directly in the css, but I believe this is not a good solution at all

tinyoverflow:
You can simply install the Tabler Core package and then include it inside the CSS and JS file.

```
npm install --dev @tabler/core 
```

```js
// resources/js/app.js
import '@tabler/core';
```

```scss
// resources/css/app.css
@import "@tabler/core";
```

Then simply embed Vite in your preferred way (eg. using `@vite(...)` in Blade) and start your Vite server with `npm run dev`.

corean:
```
// resources/js/app.js
import '@tabler/core/src/js/tabler.js';

// resources/sass/app.scss
@import '@tabler/core/src/scss/tabler';
```

Dibbyo456:
````
 harry@Harrys-MacBook-Air  ~/c-market  pnpm build

> @ build /Users/harry/c-market
> vite build

vite v4.1.1 building for production...
✓ 17 modules transformed.
transforming (18) node_modules/.pnpm/lodash.isequal@4.5.0/node_modules/lodash.isequal/index.js12:50:38 PM [vite-plugin-svelte] dom compile done.
package                 files     time     avg
@inertiajs/svelte           4   45.2ms  11.3ms
$unknown                    1   13.7ms  13.7ms
[vite:css] [sass] Can't find stylesheet to import.
  ╷
1 │ @import "../node_modules/bootstrap/scss/bootstrap";
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  node_modules/.pnpm/@tabler+core@1.0.0-beta17/node_modules/@tabler/core/src/scss/_config.scss 1:9  @import
  node_modules/.pnpm/@tabler+core@1.0.0-beta17/node_modules/@tabler/core/src/scss/_core.scss 1:9    @import
  node_modules/.pnpm/@tabler+core@1.0.0-beta17/node_modules/@tabler/core/src/scss/tabler.scss 1:9   @import
  resources/scss/app.scss 5:9                                                                       root stylesheet
file: /Users/harry/c-market/resources/scss/app.scss
Error: Can't find stylesheet to import.
  ╷
1 │ @import "../node_modules/bootstrap/scss/bootstrap";
[...]

PR #1477, opened 2023-02-20
Fix bootstrap imports

SyntaxJoker:
Related PR:

- https://github.com/tabler/tabler/pull/1384
- https://github.com/tabler/tabler/pull/1333
- https://github.com/tabler/tabler/pull/1466
- https://github.com/tabler/tabler/pull/1457

Related issues:

https://github.com/tabler/tabler/issues/1475
https://github.com/tabler/tabler/issues/1456
https://github.com/tabler/tabler/issues/1394


codecalm:
there is a bug with import stylesheet:

```
Error: Can't find stylesheet to import.
--
21:29:43.407 | ╷
21:29:43.407 | 1 │ @import "bootstrap/scss/functions";
```

SyntaxJoker:
`@import 'node_modules/bootstrap/scss/root'` will cause issues with monorepos that hoist dependencies up.
`@import "~bootstrap/scss/root"` will require the peer project to use sass-loader or something else that supports it
`@import "bootstrap/scss/root"` requires to add node modules to includePaths options

i used includePaths solution since it should resolve all reported issues

rjd22:
@SyntaxJoker thank you! This looks good to me :+1:

rjd22 (approved):

kevinpapst (changes requested):
Thank you! I am using node-sass with webpack and these changes work without any required change on my end 👍 

But I have one remark: we should use the same syntax everywhere. 
There are at least three places where still `@import "~` is used.
Could you adapt them as well @SyntaxJoker ?

- https://github.com/tabler/tabler/blob/dev/src/scss/vendor/_jsvectormap.scss#L1
- https://github.com/tabler/tabler/blob/dev/src/scss/vendor/_tom-select.scss#L5
- https://github.com/tabler/tabler/blob/dev/src/scss/vendor/_nouislider.scss#L1

SyntaxJoker:
> Thank you! I am using node-sass with webpack and these changes work without any required change on my end 👍
> 
> But I have one remark: we should use the same syntax everywhere. There are at least three places where still `@import "~` is used. Could you adapt them as well @SyntaxJoker ?
> 
> * https://github.com/tabler/tabler/blob/dev/src/scss/vendor/_jsvectormap.scss#L1
> * https://github.com/tabler/tabler/blob/dev/src/scss/vendor/_tom-select.scss#L5
> * https://github.com/tabler/tabler/blob/dev/src/scss/vendor/_nouislider.scss#L1

good point, i will change those imports aswell

kevinpapst (approved):
Thanks @SyntaxJoker 🚀

kicaj:
I confirm, thanks!

SyntaxJoker:
This is another issue, its caused by exports field
for now it can be fixed by importing tabler icons from icons-webfont like that:  `@tabler/icons-webfont/../tabler-icons.scss`
we will be soon reworking the way library is being built and this will be fixed

PR #1508, opened 2023-03-14
Fix tom-select options text without space when highlighting

### LINKED ISSUE #1505: [BUG] tom-select: Options text without space when highlighting a whole word
The texts of the options appear without space when they are highlighted according to the search text, it happens when the option has more than one word and the first word is fully highlighted.

**Screenshots**

elon

blasferna:
When a whole word or the last few letters are highlighted, the words appear without space and it doesn't look very good. With a small change applied in this pull request the bug is fixed.

fix #1505

### DIFF
--- a/src/pages/_includes/ui/select.html
+++ b/src/pages/_includes/ui/select.html
@@ -66,7 +66,7 @@
 		window.TomSelect && ({% if jekyll.environment == 'development' %}window.tabler_select["select-{{ id }}"] = {% endif %}new TomSelect(el = document.getElementById('select-{{ id }}'), {
 			copyClassesToDropdown: false,
 			dropdownClass: 'dropdown-menu ts-dropdown',
-			optionClass:'dropdown-item',
+			optionClass:'dropdown-item d-block',
 
 			{% unless include.show-search %}
 			controlInput: '<input>',


blasferna:
Hi @kevinpapst, thanks for such a quick response.  

The issue is in the "Form Elements" page.

**Before**

imagen

**After**

imagen

kevinpapst:
Ah, I see. In my project I initialize Tomselect without `dropdownClass` and `optionClass`. 
So this might be the reason why it works:

Bildschirm­foto 2023-03-14 um 11 51 56

When I change the classes on the [demo page](https://preview.tabler.io/form-elements.html) to
```
    dropdownClass: 'ts-dropdown',
    optionClass: 'option',
```
which is the default Tomselect setting, it works without `d-block`. Can you try that?

kevinpapst:
@codecalm can you remember why you are using these classes for Tomselect instead of the default ones? The default classes work perfectly across dozens of dialogs/modals/forms in my project. Shall we adapt the demo?

blasferna:
Using the following configuration I see that not all the examples look the same on the [demo page](https://preview.tabler.io/form-elements.html).

```
    dropdownClass: 'ts-dropdown',
    optionClass: 'option',
```

imagen

jozefrebjak:
> @codecalm can you remember why you are using these classes for Tomselect instead of the default ones? The default classes work perfectly across dozens of dialogs/modals/forms in my project. Shall we adapt the demo?

@kevinpapst Because they works also in dark mode, if you don't specify  `dropdownClass: "dropdown-menu ts-dropdown"` and `optionClass: "dropdown-item"` then it's works with tom-select default one, but when you switch to dark-mode then dropdown is using still the tom-select default one. @blasferna thanks, this PR solves this issue also for me.

kevinpapst:
Thanks @jozefrebjak 👍 

Now I remember, I was using Tomselect long before Tabler switched to it. 
And my setup is different from the one that @codecalm did.
I have this in my SASS build:

```
@import "tom-select/dist/scss/tom-select.bootstrap5";

@include dark-mode {
    /* selectpicker.scss / Tom Select */
    .ts-wrapper.form-select .ts-control,
    .ts-wrapper.form-select.single.input-active .ts-control {
        color: $dark-mode-text;
    }
    .ts-dropdown {
        background-color: $dark-mode-lighten;
        color: $dark-mode-text;
    }
    .ts-control {
        > input {
            min-width: unset;
            color: $dark-mode-text;
        }

        .item {
            background-color: $dark;
            color: $dark-mode-text;
        }
    }
    .optgroup-header {
        background-color: $dropdown-dark-bg;
        color: $dark-mode-text;
    }
}
```
Which looks different:
Bildschirm­foto 2023-03-16 um 15 13 27

... than Tabler default:
Bildschirm­foto 2023-03-16 um 15 17 37

Sorry, totally forgot about that. The PR looks good, even though I would vote for going the standard way and ditch the custom classes. But that is another story and should not interfere with the fix provided by @blasferna

codecalm:
@blasferna @kevinpapst I've removed Bootstrap classes from `tom-select` in https://github.com/tabler/tabler/commit/734633200ac82a1b4437d7a5aef9a8fd6f79d5d8 :) Now it works much better

Issue #1525, opened 2023-04-09
border width classes do not work (e.g. border-1, border-2, border-3, etc)


nbrookie:
Example:

```
<div class="border border-4">
  Banana
</div>
```

matches the following css for the banana div:
```
.border-4 { 
  --tblr-border-width: 4px;
} 
.border { 
  border: 1px var(--tblr-border-style) rgba(97,104,118,.16)!important
} 
```

Notice how the `border` property has a hard coded width of `1px` and does not take into consideration the `--tblr-border-width` variable?

PR #1526, opened 2023-04-09
Fix bug where border-1, border-2, etc don't work

nbrookie:
Fixes https://github.com/tabler/tabler/issues/1525


codecalm:
thank you for your help! ❤️

Issue #1536, opened 2023-04-19
[BUG] Checkbox `:indeterminate` CSS color


cavasinf:
**Describe the bug**

Checkbox CSS of `:indeterminate` state is not linked to the CSS tabler vars.
https://github.com/tabler/tabler/blob/e84ab993119879d26a9b9c30095e8a0a2bab8abd/dist/css/tabler.css#L2870-L2874

It should be like `:checked` behavior
https://github.com/tabler/tabler/blob/e84ab993119879d26a9b9c30095e8a0a2bab8abd/dist/css/tabler.css#L2860-L2863

**To reproduce**

1. Override `--tblr-primary` color
2. Add checkboxe
3. Set `indeterminate` to `true`
4. See wrong color

**Screenshots**
![image](https://user-images.githubusercontent.com/25293190/233008040-bbfdf2cb-60d1-4b92-8cac-4be31f7e97a9.png)

codecalm:
@Saran-pariyar we should add to config:

```scss
$form-check-input-indeterminate-bg-color: var(--#{$prefix}primary) !default;
```

PR #1548, opened 2023-04-28
Fixed issue #1536

Saran-pariyar:


src/scss/_variables.scss
```diff
@@ -882,6 +882,7 @@ $form-feedback-icon-invalid: str-replace(
 
 $form-label-font-size: $h4-font-size !default;
 $form-label-font-weight: var(--#{$prefix}font-weight-medium) !default;
+$form-check-input-indeterminate-bg-color: var(--#{$prefix}primary) !default;
```
codecalm, line 885: better way is to add it to `$form-check-input-XXX` section (around lines 852)

codecalm (changes requested):

src/scss/_variables.scss
Saran-pariyar: Placed it in line no. 852 in the 2nd commit

Issue #1513, opened 2023-03-23
css variable without value in popover


jeijei4:
https://github.com/tabler/tabler/blob/eeb1a925e0b12c561609f4e813e758f92f9b4a3a/dist/css/tabler.css#L6652

Other remarks:

- The following line numbers in the <code>tabler.css</code> file have **duplicate properties**:
44, 46-51, 53-55, 154, 177, 180, 204, 

- line number 236: "Cannot resolve '--tblr-body-text-align' custom property".

- line number 366: "Cannot resolve '--tblr-link-opacity' custom property".

- line number 413: "Cannot resolve '--tblr-muted-dark' custom property".

- line number 779: "Cannot resolve '--tblr-rows' custom property".

- line number 780: "Cannot resolve '--tblr-columns' custom property".

- line number 781: "Cannot resolve '--tblr-gap' custom property".

- line number 3796: "Cannot resolve '--tblr-dropdown-item-border-radius' custom property".

- line number 3951: "css variable '--tblr-nav-link-font-weight' without value".

- line number 3965: "Cannot resolve '--tblr-nav-link-font-size' custom property ".

- line number 4196: "Cannot resolve '--tblr-scroll-height' custom property ".

- line number 4500: "css variable '--tblr-card-title-color' without value".

- line number 4501: "css variable '--tblr-card-subtitle-color' without value".

- line number 4511: "css variable '--tblr-card-height' without value".

- line number 4838: "Cannot resolve '--tblr-breadcrumb-font-size' custom property ".

- Unable to resolve multiple custom properties, from line 5074 to 5266 and 5690 to 6106 .

- line number 6162: "css variable '--tblr-toast-color' without value".

- line number 6229: "css variable '--tblr-modal-color' without value".

- line number 6243: "css variable '--tblr-modal-footer-bg' without value".

- line number 6521: "css variable '--tblr-tooltip-margin' without value".

- line number 6538: "Property text-align overwrites property text-align ".

- line number 6632: "css variable '--tblr-popover-header-color' without value".

- line number 6648: "Property text-align overwrites property text-align ".

- line number 10271: "Cannot resolve '--tblr-emphasis-bg-rgb' custom property ".

rjd22:
Closing as a duplicate of #1490

PR #1549, opened 2023-04-30
#1513 issue:updated css varible without value in popover.

Nitz2611:


kevinpapst:
You edited the files in `dist/` and `demo/dist/` directory. Your changes would be overwritten on the next build. Please apply your changes in the source files instead.

Nitz2611:
@kevinpapst there is no tabler.css in source file. but in issue it says --tblr-popover-header-color variable is without value in [tabler/dist/css/tabler.css]

kevinpapst:
I assume that this is a (new) bootstrap variable that needs to be defined.
Search for `$popover-header-color` in `node_modules/bootstrap/scss/_variables.scss`

The relevant parts are:
```
$headings-color:              null !default;
$popover-header-color:              $headings-color !default;
```

Interesting is, that even bootstrap itself seems to have that bug: 
Bildschirm­foto 2023-05-01 um 14 23 01

No clue what to do though

Nitz2611:
So basically it is bootstrap bug. There is nothing we can do about it. Right?

kevinpapst:
Initializing the variable with null looks like a bootstrap issue, yes.
Better if @codecalm will have a look at that.

PR #1575, opened 2023-05-15
Add new color picker component using `coloris.js` library

codecalm:
This PR adds a new Color Picker component to our project, using the `coloris.js` library. The new component provides users with a customizable color picker, allowing them to select and preview colors for use in their projects.

Demo: https://tabler-git-dev-colorpicker-tabler-io.vercel.app/colorpicker.html

### Preview

Screenshot 2023-05-15 at 03 00 48
Screenshot 2023-05-15 at 03 00 55


kevinpapst:
Only one problem: 
It seems to work perfect with the keyboard, but how does one close that dialog with the keyboard? 
Seems ESC closes it without using the chosen color.

diomed:
> how does one close that dialog with the keyboard?

I posted your in issues of colorpicker on github, since that's their problem.. so we'll see.

diomed:
@kevinpapst developer said TAB + ENTER does the trick on keyboard

kevinpapst:
Thanks for asking! Just tested again: partially correct. The focus has to be on this field and then enter closes the dialog.
Bildschirmfoto 2023-05-16 um 13 23 20
Counter intuitive... but at least it works ^^

diomed:
Yes. unsure why you'd need that. for handheld devices maybe? when no mouse is available.

Maybe it would be a good idea to also write a quick little nice info about that on colorpicker.html page
instead of lorem ipsum text

codecalm:
@kevinpapst I've added fixes to focus element

PR #1593, opened 2023-05-23
Adding Two-Step Verification Pages

codecalm:
localhost_3000_2-step-verification html_

localhost_3000_2-step-verification-2 html_

### DIFF
--- a/.changeset/fluffy-insects-lay.md
+++ b/.changeset/fluffy-insects-lay.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Adding Two-Step Verification Pages
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,7 +1,7 @@
 {
 	"bracketSpacing": true,
 	"jsxSingleQuote": false,
-	"printWidth": 80,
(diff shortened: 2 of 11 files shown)


kevinpapst:
I like the design ❤️ could be used for 2FA / totp token screens as well!

Currently I am doing TOTP in one single input field to get around the javascript handling of jumping between the inputs.
I think this is missing here. It looks nice (no question!), but most users don't properly use their keyboard, let alone the TAB key. They will have to use the mouse to navigate between the fields, and that is ... well no the best UX.

I think keyboard handling (typing, backspace, paste from clipboard) should be added in this screen to make it complete.

codecalm:
@kevinpapst i've added JavaScript interactions! 🙂 

https://github.com/tabler/tabler/assets/1282324/0189041f-91f3-4ab5-9b23-fea8f424a60d

kevinpapst:
I li-la-love it, thanks @codecalm 👍 
Icing on the cake wishes:
- autofocus of the first field upon page load 
- support for pasting a code (currently always adds the first character to one field)

Anyway: I love it already and can't wait to use it 😁

Issue #1588, opened 2023-05-21
[BUG] Important alert titles have the same color as their background


lovasoa:
**Describe the bug**

An `alert-title` inside an `alert-important` renders the alert title unreadable.

![image](https://github.com/tabler/tabler/assets/552629/b20cd40c-f4f3-4593-a4db-5fee5a97199d)

```html
<div class="
    alert
    alert-red
    alert-dismissible
    alert-important
" role="alert">
  <div class="d-flex">

    <div class="icon alert-icon">
      <i class="ti ti-alert-circle"></i>
    </div>

    <div>
      <h4 class="alert-title">Alert</h4>
      <div class="alert-description">SQLPage is entirely free and open source.</div>
          
    </div>

  </div>
    <a class="btn-close" data-bs-dismiss="alert" aria-label="close"></a>
</div>
```

Nitz2611:
Can you assign me fo this?

BamaCharanChhandogi:
Hey @lovasoa, This code is in which folder?

lovasoa:
The snippet of code above was generated by sqlpage, it's not in any codebase.

lovasoa:
I'm not sure I understand the question. The snippet of html in the original post reproduces the problem. The problem is that `alert-important` changes the background color, but not the color of the alert title, that becomes invisible.

Nitz2611:
alert-title and alert-important both inherit same color from alert-red, thats why this problem occurs.

Nitz2611:
#1610

lovasoa:
thank you !

allejo:
Was this fix pushed out in a new release and/or pushed to the documentation website?

<img width="722" alt="image" src="https://github.com/tabler/tabler/assets/1246453/d6c62f82-af3e-4de3-9501-67e7723177cf">

PR #1609, opened 2023-05-31
fix: #1588 add important alert with title

rjd22:
@codecalm I'm not sure I did this right. So much has changed in the last versions. What is the correct way to adapt the docs?

I also wanted to fix the issue with the docs hiding part of the alerts but was unable to understand what is used to generate them:

2023-05-31T10:10:25,357014965+02:00

### DIFF
--- a/.changeset/hip-jobs-double.md
+++ b/.changeset/hip-jobs-double.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": minor
+---
+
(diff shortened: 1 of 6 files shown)


codecalm:
@rjd22 everythink looks perfect! I added some examples to `alerts.html` page and added more examples. 

https://tabler-git-fix-alert-titles-not-visible-tabler-io.vercel.app/alerts.html

PR #1623, opened 2023-06-11
Unify size of avatar, flag and payment components

codecalm:


codecalm:
Screenshot 2023-06-12 at 01 33 53

PR #1645, opened 2023-07-08
Fix missing tfoot styling

promatik:
`tfoot` doesn't have the same behaviour as `thead`, leading to this behaviour;

image

This PR aims to fix it 👌

### DIFF
--- a/src/pages/_includes/ui/table.html
+++ b/src/pages/_includes/ui/table.html
@@ -86,5 +86,25 @@
 		</tr>
 		{% endfor %}
 		</tbody>
+		{% if include.footer %}
+		<thead>
+			<tr>
+				<th>Name</th>
+				<th>Title</th>
+	
+				{% unless include.avatars %}
+				<th>Email</th>
+				{% endunless %}
+	
+				<th>Role</th>
+	
+				{% if include.nowrap %}
+				<th></th>
+				{% endif %}
+	
+				<th class="w-1"></th>
+			</tr>
+		</tfoot>
+		{% endif %}
 	</table>
 </div>
--- a/src/pages/_includes/ui/table.html
+++ b/src/pages/_includes/ui/table.html
@@ -87,7 +87,7 @@
 		{% endfor %}
 		</tbody>
 		{% if include.footer %}
-		<thead>
+		<tfoot>
 			<tr>
 				<th>Name</th>
 				<th>Title</th>
--- a/src/pages/tables.html
+++ b/src/pages/tables.html
@@ -6,7 +6,7 @@ menu: base.tables
 <div class="row row-cards">
 	<div class="col-lg-8">
 		<div class="card">
-			{% include ui/table.html card=true limit=8 %}
+			{% include ui/table.html card=true limit=8 footer=true %}
 		</div>
 	</div>
 
--- a/src/pages/tables.html
+++ b/src/pages/tables.html
@@ -6,7 +6,7 @@ menu: base.tables
 <div class="row row-cards">
 	<div class="col-lg-8">
 		<div class="card">
-			{% include ui/table.html card=true limit=8 footer=true %}
+			{% include ui/table.html card=true limit=8 %}
 		</div>
 	</div>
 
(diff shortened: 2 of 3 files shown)


kevinpapst:
Can you share before and after images of a complete table?
Unfortunately the table demo https://tabler-git-fork-promatik-thead-tfoot-tabler-io.vercel.app/tables.html does not contain any `<tfoot>`, so it's not possible to compare the change visually with the Vercel deployment.

Is your given screenshot the current state (as in "before")? 
Because it looks like that and I would say: it's correct. My use-case is usually: `<tfoot>` contains totals of the above table and those need to be bold and in a regular font-size. So it contains content and should be typographically treated like other (table) content.

To me, this looks wrong:
Bildschirmfoto 2023-07-09 um 02 00 22 
I would appreciate if this styling would not become behavior of tabler, but opt-in via some class.

This example is from the [official bootstrap docs](https://getbootstrap.com/docs/5.0/content/tables/#table-foot):
Bildschirmfoto 2023-07-09 um 02 03 06
Visually `<tfoot>` has no difference to text in `<tbody>`.

promatik:
Hi @kevinpapst!
Thank you for quick response.

Here is a before;
image

And after;
image

I was not expecting someone wanted a different behaviour/theme between `<thead>` and `<tfoot>` 😅

We use `<tfoot>` for guidance, if you're checking a cell near the bottom, is easier to get the guidance from there, and in that way, we want the layout to be the same.
I'm introducing a fix at [backpack](https://backpackforlaravel.com/) for this, so if this one doesn't get merged it's ok, but personally, me as a developer, I would expect this behaviour 👌

promatik:
Also, added a footer to the **second** table on demo;
https://tabler-git-fork-promatik-thead-tfoot-tabler-io.vercel.app/tables.html

kevinpapst:
> We use <tfoot> for guidance, if you're checking a cell near the bottom, is easier to get the guidance from there, and in that way, we want the layout to be the same.

I am using `<thead class="sticky-top">` therefor, so the column names are always visible.

promatik:
> _Have you tried using <thead class="sticky-top"> for that?_

Yes, we have that too 😅 it's really for a matter of guidance/accessibility.

But I understand your point, most probably since this is a requisite on our side we should do the dirty job, for now it's done, we have the css override for `<tfoot>`.
I'll let you guys decide it 👌

rjd22:
I rather not merge stuff that is off spec and user specific. @codecalm do you have any thoughts on this?

kevinpapst:
@codecalm That is a huge BC break for every user. Now we have to add rules to fix a broken Table. Did you read the provided links?

codecalm:
I've checked demos from this PRs, so for me it's okay to add this feature to tabler

codecalm:
I want to rebuild the tables decently in the future, in the latest version of bootstrap there will also be a big refactoring there, so I think in this form it can go for the moment

rjd22:
@codecalm our point is that tfoot is meant for summaries of the tables like totals. That is why they are bold. 

Changing the sematic nature might break existing installs that are using tfoot correctly.

codecalm:
@rjd22 sure, I will check how it is solved in other libraries and make a new PR with corrections

rjd22:
Thank you! It's not that I'm against this functionality but it's better to move it outside the default with a class like `tfoot.column-names`. This way we can introduce this functionality without breaking the original.

kevinpapst:
Here are some links to some frameworks: 

- Tailwind: https://flowbite.com/docs/components/tables/#table-foot
- Bulma: https://bulma.io/documentation/elements/table/
- Bootstrap: https://getbootstrap.com/docs/5.2/content/tables/#table-foot

I think seeing Tailwind and Bootstrap alone should be enough reasoning to revert.

But I repeat my quote of the HTML5 spec, in case you missed it before:

> The tfoot element represents the block of rows that consist of the column summaries (footers) for the parent table element

In other words: the table footer IS NOT the table header. The example at https://tabler-git-fork-promatik-thead-tfoot-tabler-io.vercel.app/tables.html which was added by @promatik is highly misleading as this is NOT what a `<tfoot>` should be used for. The example should be replaced by a real world scenario of a regular data table that contains summaries in tfoot.

Please make it opt-in with a new `.table-tfoot-like-thead` class. 🤷 

And another btw: we gave a lot of comments before which were simply ignored(?) That feels disappointing from a collaboration point of view.

PR #1646, opened 2023-07-08
Fix Dockerfile

mattsmithdev:
Update Dockerfile to remove/add proper ADD commands and include the correct packages to be able to build/run


rjd22:
Looks good but from some reason the build fails. @codecalm can you look at this when you have the time?

codecalm:
@mattsmithdev thank you for your help! 🙂

PR #1647, opened 2023-07-09
Avoid scss color dependency on :focus

promatik:
As of now, focus border and shadow color depend on scss variables, therefore tabler compilation.
IMO it shouldn't, actually nothing should depended on scss variables, specially colors.

This PR allows inputs border and shadow to depend on primary color css variable, devs will be able to do this;

image

### DIFF
--- a/src/scss/mixins/_mixins.scss
+++ b/src/scss/mixins/_mixins.scss
@@ -63,6 +63,6 @@
   box-shadow: 0 0 $focus-ring-blur $focus-ring-width rgba(var(--tblr-primary-rgb), 0.25);
 
   @if($show-border) {
-    border-color: rgba(var(--tblr-primary-rgb), 0.5);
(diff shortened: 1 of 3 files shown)


src/scss/mixins/_mixins.scss
```diff
@@ -60,9 +60,9 @@
 
 @mixin focus-ring($show-border: false) {
   outline: 0;
-  box-shadow: $focus-ring-box-shadow;
+  box-shadow: 0 0 $focus-ring-blur $focus-ring-width rgba(var(--tblr-primary-rgb), 0.25);
 
   @if($show-border) {
-    border-color: $focus-ring-border-color;
+    border-color: rgba(var(--tblr-primary-rgb), 0.5);
```
promatik, line 66:
```suggestion
    border-color: rgba(var(--tblr-primary-rgb), 0.25);
```

codecalm:
thank you for your help! ❤️

PR #1675, opened 2023-08-17
Add icons-page with layout

BG-Software-BG:


dheineman:
Hello, just checking in case you are not aware but all icons are already listed and searchable on https://tabler-icons.io/

BG-Software-BG:
> Hello, just checking in case you are not aware but all icons are already listed and searchable on https://tabler-icons.io/

Hi. Yes, I'm aware of that :)

PR #1679, opened 2023-08-22
Fix hover background colors

mholt:
I am not very familiar with SCSS but I found that renaming `--tblr-text-secondary-rgb` (which is not defined) to `--tblr-secondary-rgb` made it work for me :man_shrugging:


mholt:
Closing this in favor of a new PR I started by cloning down the repo, installing Docker, spinning everything up, and making some adjustments that I think will be helpful :+1:

PR #1713, opened 2023-09-18
Fix hover background

BG-Software-BG:
Use `--tblr-secondary-rgb` variable instead of not existing `--tblr-text-secondary-rgb`;

image
->
image


rjd22:
This seems to be already fixed so closing this.

PR #1720, opened 2023-09-23
Fix replacing href="#" with href="javascript:void(0)"

BG-Software-BG:
Replaces before using example variable.
Replace all hrefs, not only in <a> tags.

Should prevent opening whole site inside iframe in examples in Documentation.


rjd22:
Thank you for the fix!

PR #1733, opened 2023-10-01
Change h1 to div in navbar-logo

BG-Software-BG:


kevinpapst (approved):
Thank you 👍

rjd22:
Looks to to me! I would suggest to also move the page titles to h2 then but we can do this in a separate PR.

PR #1745, opened 2023-10-19
Fix dynamic scss prefix in mixins

datlechin:


src/scss/mixins/_mixins.scss
```diff
@@ -60,9 +60,9 @@
 
 @mixin focus-ring($show-border: false) {
   outline: 0;
-  box-shadow: 0 0 $focus-ring-blur $focus-ring-width rgba(var(--tblr-primary-rgb), 0.25);
+  box-shadow: 0 0 $focus-ring-blur $focus-ring-width rgba(var(--#{$prefix}-primary-rgb), 0.25);
```
kevinpapst, line 63: Shouldn't that be `--#{$prefix}primary-rg` instead?

datlechin: oops sorry, my bad

rjd22:
Looks good to me 👍

PR #1794, opened 2023-12-09
Revert "Fix missing tfoot styling"

codecalm:
Reverts tabler/tabler#1645


kevinpapst:
Thank you @codecalm - that was the right move imo 👍

Issue #1768, opened 2023-11-10
[BUG] Use of a secondary color in certain form element


cavasinf:
### Browser

Firefox

### OS

Windows 11

### Screen size

1920x1080

### Describe the bug

I don't think it is a good idea to use `--tblr-secondary` on "secondary" texts.
If our client uses 2 colors in their company color code,
we end up with some weird form input when the secondary color is overridden.

Maybe use the `muted` color as it is already done with the input placeholders
https://github.com/tabler/tabler/blob/c40ca3039423a33a693562c1fae0a15806d1a23e/site/styles/_forms.scss#L42-L44

Even the footer has this color weird thing:
![image](https://github.com/tabler/tabler/assets/25293190/4b2ac19b-b7f2-4891-afc8-16eb3c108dc3)

### How to reproduce

1. Change `--tblr-secondary` css variable color
2. See form-hint color
...

### Screenshots

![image](https://github.com/tabler/tabler/assets/25293190/9317c6d2-2e98-4b55-a6d0-2cfe805ef486)

![image](https://github.com/tabler/tabler/assets/25293190/d78ea6d4-e022-4c50-beac-055394f25442)

![image](https://github.com/tabler/tabler/assets/25293190/c7c94621-f50b-4b1c-95d4-e1f0eb0dd146)

ethancrawford:
Look like this is still an issue after theming was introduced in https://github.com/tabler/tabler/pull/2054

cavasinf:
Yes, still the case in many components/pages.
`--tblr-secondary` is still used as "default is grey so it is a good color for muted text", or it is not true.
`--tblr-muted` should be used instead @codecalm.

<img width="1313" height="651" alt="Image" src="https://github.com/user-attachments/assets/e066c7e4-8229-4e8f-913c-58233359be19" />

PR #1796, opened 2023-12-11
Fix for #1768 - Use of secondary color in certain form element

jhanavigera:
This PR aims to resolve the issue mentioned in #1768 

Steps taken to resolve the issue:
- I created a new variable called $form-secondary-color which would allow us to have a globally initialzed secondary color value for the forms.scss file. 
- I also declared this variable to be equal to the gray-500 as it was the value for the previous variable(var(--#{$prefix}secondary)). 
- I replaced the definition of var(--#{$prefix}secondary) to var(--#{$prefix}gray-500) in the _variables.scss file. 
- I noticed that on local developmental environment, some of the colours are being taken from the demo\dist\css\tabler.css file and I was wondering whether I should change the definition within this file. As based on my understanding, this file is used to deploy a specific version of the application and shouldn't be modified during development. 

This change was tested on a Windows environment using the Google Chrome browser. All variables mentioned under the src/ directory are being displayed as expected.

### DIFF
--- a/src/scss/_variables.scss
+++ b/src/scss/_variables.scss
@@ -912,6 +912,8 @@ $form-feedback-icon-invalid: str-replace(
 $form-label-font-size: $h4-font-size !default;
 $form-label-font-weight: var(--#{$prefix}font-weight-medium) !default;
 
+$form-secondary-color: var(--#{$prefix}gray-500) !default;
+
 // Legend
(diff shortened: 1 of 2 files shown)


jhanavigera:
@codecalm can you please have a look at my PR when you get a chance. Let me know if any changes are required and I'll be more than happy to make them. This is my first ever contribution to open source and I am really looking forward to your input!

Issue #1397, opened 2022-12-04
[BUG] Bottom border in Card Table is too thick


tinyoverflow:
**Describe the bug**

When using a table inside a card, the bottom border is too thick, as it shows the bottom border of the table and of the card.

**Device**

Device independent.

**To reproduce**

Use a default card with a table inside it. Give the table the `card-table` class:

<details>

<summary>View Example Code</summary>

```html
<div class="card">
    <table class="table card-table">
        <thead>
            <tr>
                <th>Name</th>
                <th>Title</th>
                <th>Email</th>
                <th>Role</th>
                <th class="w-1"></th>
            </tr>
        </thead>
        <tbody>
            <tr>
                <td>Kellie Skingley</td>
                <td>Teacher, Services</td>
                <td>kskingley7@columbia.edu</td>
                <td>Owner</td>
                <td><a href="#">Edit</a></td>
            </tr>
        </tbody>
    </table>
</div>
```

</details>

**Screenshots**

![grafik](https://user-images.githubusercontent.com/6287093/205486872-e35827b8-9fb2-4a52-8646-7a713400c5d8.png)

Abhishek2nayak:
@tinyoverflow i would like to contribute on this issue.

tinyoverflow:
Okey, just create a PR if you want. I don't have time for this at the moment, unfortunately.

rjd22:
This seems to be solved now so closing this.

tinyoverflow:
@rjd22 As far as I can tell, this hasn't been fixed. At least not in the main branch. What PR should've fixed this?

Saran-pariyar:
the bottom-border is a little thick in some places:
<img width="292" alt="image" src="https://github.com/tabler/tabler/assets/86418083/bc7c96aa-73ba-4071-8c93-b150dd833dd2">


and in that places, the table already has class name "card-table":
<img width="303" alt="image" src="https://github.com/tabler/tabler/assets/86418083/111d407a-ebd2-47cf-87bf-45b968fb27ba">



so in that place, i made the card bottom border =0, and it worked cause now only the table border is there. Will this do it?

<img width="300" alt="image" src="https://github.com/tabler/tabler/assets/86418083/3e5f91a1-715e-45bc-8787-90c942318a66">
 
<img width="299" alt="image" src="https://github.com/tabler/tabler/assets/86418083/7d60d487-0525-45c7-84fd-2921312e0729">

tinyoverflow:
[...]

PR #1826, opened 2024-02-12
fixing "[BUG] Bottom border in Card Table is too thick #1397"

resul-elezi:


kevinpapst (changes requested):
You changed the dist files, but we need the change in the source scss files:
https://github.com/tabler/tabler/tree/dev/src/scss

PR #1859, opened 2024-03-30
Update layout-rtl.html

UnrivaledIr:
Enhanced RTL Demo: Replaced English Data with Farsi for Clarity

### DIFF
--- a/demo/layout-rtl.html
+++ b/demo/layout-rtl.html
@@ -12,7 +12,7 @@
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover"/>
     <meta http-equiv="X-UA-Compatible" content="ie=edge"/>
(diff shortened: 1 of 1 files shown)


kevinpapst (changes requested):
This does not work unfortunately.
The idea is good to use RTL content to show the RTL support.
But you change the generated HTML file and not the source file, your changes would be lost with the next update. The source file is here: https://github.com/tabler/tabler/blob/dev/src/pages/layout-rtl.html

And I don't think it can be fully translated without a lot of changes for this single use-case.

You would have to traverse through all Jekyll templates, layout and include files and make them translatable somehow.

What you could do is to change the `page-header` and `page-title` variables in the linked file.

kevinpapst:
Thanks for taking your time to submit your PR. Unfortunately can't merge as is.
Closing this stale issue. Please re-open with the explained changes.

PR #1885, opened 2024-05-02
Fixes #1820

### LINKED ISSUE #1820: [BUG] Terms of Service demo page is cut off
### Browser

Chrome 120

### OS

Win 10

### Screen size

1920x1080

### Describe the bug

The top side of the page https://preview.tabler.io/terms-of-service.html is cut off (item 1 is not visible at all)

### How to reproduce

1. Go to https://preview.tabler.io/terms-of-service.html
2. top of the page is missing, article 1 not visible

### Screenshots

_No response_

### JSFiddle

_No response_

bjorntheart:


kevinpapst (changes requested):
Thank @bjorntheart 👍 

Now could you please make the change in the src file and not the generated HTML.
The source file is here: https://github.com/tabler/tabler/blob/dev/src/pages/terms-of-service.html#L7

PR #1925, opened 2024-07-14
[modals]: style btn-close only when in .modal-header

blikblum:
Currently if you put a .btn-close element in Modal body, the element is displayed in top right.

See below how  the below code (a .tag with a .btn-close) is rendered  (look at the smaller close below the header close):

```html
<div class="modal-body">
  <span class="tag"> Test <button type="button" class="btn-close"></button></span>
</div>

```
image

This PR fixes it

## Summary by CodeRabbit

- **New Features**
	- Improved modal layout with updated styling for better usability.
	- Enhanced specificity of the close button positioning within modals.

### DIFF
--- a/src/scss/ui/_modals.scss
+++ b/src/scss/ui/_modals.scss
@@ -1,4 +1,4 @@
-.modal-content {
+.modal-header {
   .btn-close {
     position: absolute;
     top: 0;
--- a/src/scss/ui/_modals.scss
+++ b/src/scss/ui/_modals.scss
@@ -1,4 +1,4 @@
-.modal-container, .modal-header {
+.modal-content, .modal-header {
   > .btn-close {
     position: absolute;
     top: 0;
--- a/src/scss/ui/_modals.scss
+++ b/src/scss/ui/_modals.scss
@@ -1,5 +1,5 @@
-.modal-header {
-  .btn-close {
+.modal-container, .modal-header {
+  > .btn-close {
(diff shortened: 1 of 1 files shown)


rjd22:
@blikblum AFAIK this is expected behaviour. As not all modals and card have headers. What problem are you trying to solve?

blikblum:
Put a [tag component](https://preview.tabler.io/tags.html) inside a modal:

```html
<div class="modal-body">
  <span class="tag"> Test <button type="button" class="btn-close"></button></span>
</div>
```

The close button of all tags will be displayed in top right corner. See image in OP

blikblum:
> As not all modals and card have headers.

Original bootstrap code only style close button inside header. To support close btn outside a header, IMO tabler should do using a dedicated markup

rjd22:
In that case I would change it that it needs to be a direct descendant of modal-header or modal-content. That way both situations will work.

src/scss/ui/_modals.scss
```diff
@@ -1,5 +1,5 @@
-.modal-content {
-  .btn-close {
+.modal-container, .modal-header {
```
rjd22, line 1: Shouldn't the first be .modal-content?

blikblum: yep. Fixed

blikblum:
@rjd22 many thanks

PR #1965, opened 2024-09-24
Add the possibility to include Tabler as a Composer package

dpoposki:
I would like to add the possibility to inlcude Tabler as a Composer package, so that PHP projects would be able to include it as a vendor and be able to use the SCSS or JS code.


dpoposki:
Seems this is not possible due to the package expecting that bootstrap is there via the package.json dependency.

PR #1995, opened 2024-11-21
FIX: ensure border color works in dark mode

L0n35:
This fix was needed because in dark mode the border color of these utility classes was not changing to its dark mode equivalent.


plaimbock:
Thank you for your contribution!

PR #2006, opened 2024-12-04
Added accent color classes

p-fazzini:
Added classes to set the accent color of various elements based on the data-attribute added in the html tag


codecalm:
hi @p-fazzini, thank you for your contribution!

the same results you can do with `--tblr-primary: green;` in `html` selector. Overriding those css are not necessary

p-fazzini:
@codecalm Certainly, I am aware of that! However with modification I introduced the end user can decide his favorite theme color and each user can have his own color scheme

codecalm:
we've created new props in https://github.com/tabler/tabler/pull/2054

PR #2008, opened 2024-12-05
Replaced TinyMCE with Hugerte to address license violation

### LINKED ISSUE #1983: [LICENSE VIOLATION] TinyMCE copyleft license FORBIDS this to be licensed MIT
### Browser

does not matter

### OS

does not matter

### Screen size

does not matter

### Describe the bug

TinyMCE is licensed GPLv2+ since the 7.x version, making it incompatible with permissively-licensed open source products like this one. Therefore, TinyMCE must be removed or Tabler switch to GPL, which would be really bad. Instead, I recommend you replacing TinyMCE with my [HugeRTE](https://github.com/hugerte/hugerte) fork. Thank you!

### How to reproduce

see above

### Screenshots

_No response_

### JSFiddle

_No response_

codecalm:
Fixes #1983


kevinpapst:
If you stick to the old TinyMCE 6, there shouldn't be a problem.
Even more so, if you use it only via CDN and do not depend on it directly.

I personally have no use for TinyMCE and never used it. 
So if it gets unbundled from Tabler, it would be even better (for me).

This might be a better way anyway for the future: to build small integration packages instead of a huge build that includes all these choices that the projects which use Tabler might not need.
There could be packages for
- TomSelect
- FullCalendar
- TinyMCE
- ...

rjd22:
@kevinpapst As far as I know TinyMCE is not bundled with Tabler. It's an peerDependency so it don't be installed if you're not using it.

rjd22:
> The TinymCE (or all other tool) specific SCSS gets compiled into the full bundle, which is not necessary for many use-cases. The browser world looks different today and having a huge build is not neccessary anymore. We can have many small (component) CSS files instead.

I don't disagree with you, but I would argue that bundle size and distribution is something that users need to control. The reason is that there are so many different ways to use a theme that the theme developer cannot take into account every way of usage.

IMO I think it's okay to ship a bundle for the download so that users can get that with batteries included. For the more power users that want to strip it apart (Like I did). I recommend downloading it trough NPM and copy the core file (https://github.com/tabler/tabler/blob/dev/src/scss/_core.scss) and only include what they need.

They could even use the separate ui module files and include them in their components when using something like React or VueJS.

carlosmintfan:
> If you stick to the old TinyMCE 6, there shouldn't be a problem.

But if you use HugeRTE you'll slowly get improvements while TinyMCE 6 is EOL.

rjd22:
> > If you stick to the old TinyMCE 6, there shouldn't be a problem.
> 
> But if you use HugeRTE you'll slowly get improvements while TinyMCE 6 is EOL.

The problem lies in trust. We cannot just include libraries from everyone because tabler is a really big project and we have to deal with the thought that bad actors would like their library included in big projects like tabler.

I'm not saying you're a bad actor but I also have no way of verifying that you're not.

For that reason @codecalm let's not change the wysiwyg editor just yet and wait and watch how the project is evolving.

We're not breaking any TinyMCE license atm as far as I know.

kevinpapst:
> We're not breaking any TinyMCE license atm as far as I know

This and Tabler is not bundling or linking TinyMCE. 
AFAIK the GPL only applies if you link against the package in the distribution.
And here it is only used in the demo pages, which are not part of the actual NPM package.

carlosmintfan:
If you should decide to merge this one day (this has not been closed yet), please capitalize `rte` for `HugeRTE` instead of `Hugerte`, like you do for `TinyMCE`. :smile:

preview/package.json
```diff
@@ -41,17 +41,18 @@
...
+    "hugerte": "^1.0.9",
     "imask": "^7.6.1",
     "jsvectormap": "^1.6.0",
     "list.js": "^2.3.1",
     "litepicker": "^2.0.12",
     "nouislider": "^15.8.1",
     "plyr": "^3.7.8",
+    "signature_pad": "^5.0.6",
     "star-rating.js": "^4.3.1",
     "tinymce": "^7.7.1",
```
kevinpapst, line 53: tinymce is still required

codecalm: fixed

PR #2021, opened 2024-12-31
Make horizontal rule direction aware

### LINKED ISSUE #1510: [FEATURE] make hr-text-left/right direction aware
**Is your feature request related to a problem? Please describe.**

https://preview.tabler.io/docs/divider.html

When using RTL languages, the classes `hr-text-left` and `hr-text-right` are troublesome.
They do not switch their direction, but always align as they are named, right or left.

When using the rtl CSS I would like to see the divider switch their direction automatically.

**Describe the solution you'd like**

I would like to see new classes that use the same naming and logic as Bootstrap for directions: 
`hr-text-start` and `hr-text-end`.

For BC sake, new classes should be added, while the old ones retaining their current logic. 

**Describe alternatives you've considered**

The least we should do is switch directions on RTL. But that would be misleading.

codecalm:
closes #1510

### DIFF
--- a/.changeset/chilly-vans-leave.md
+++ b/.changeset/chilly-vans-leave.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Make horizontal rule direction aware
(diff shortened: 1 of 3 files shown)


kevinpapst:
Thanks 👍

Issue #1803, opened 2023-12-28
[FEATURE]  -  Improve the sidebar navigation


daraduda:
**Is your feature request related to a problem? Please describe.**

- Improved sidebar navigation in dark and light modes.

**Describe the solution you'd like**

- Please see this UI:  [Tabler sidebar navigation](https://dribbble.com/shots/21217466-Tabler-sidebar-navigation)

![nav](https://github.com/tabler/tabler/assets/27419093/9279eafe-1e1c-48a1-8147-6bee2091f13d)

daraduda:
![nav](https://github.com/tabler/tabler/assets/27419093/9279eafe-1e1c-48a1-8147-6bee2091f13d)

Retr0-dot-cpp:
could you provide with other examples because the first ui you sent is no longer available

ivalkenburg:
This would be a great addition for tablet viewport.

kevinpapst:
For reference, there is a duplicate issue that was created one year ago: #1474 

Let's keep yours as well, as it quotes the image from @codecalm 

I wonder why I can't find a post on Twitter about this and not on the Tabler blog either ...

abrahampe:
When this feature get ready I (and many others developers) will adopt tabler!

hozzaq:
it looks great, can't wait

abrahampe:
[OFF-TOPIC] Because of things like these that programmers always need partners with sales skills.

vjmartins:
I'm available to be a beta tester of the feature.. 😅

vjmartins:
Hi @BG-Software-BG ,

Any update?

BGaweda:
Work in progress. The sidebar is expected to be ready in January.

eclipsee404:
Hi @BG-Software-BG 

Any update for this feature?

BGaweda:
Still working on it

mr1716:
Keeping my eye on this as interested in completion

vrozkovec:
@BGaweda any info on this, please? I'm in the process of migrating several legacy systems, and this would come really handy.

mr1716:
@vrozkovec, the question to ask is when the Pull Request will be handled and merged in. As it may solve this issue, then hopefully it gets done soon and they work on it

BGaweda:
Unfortunately no updates on this issue

mr1716:
@BGaweda any rough timeline or timetable for this to be implemented?

bessone:
[...]

PR #2043, opened 2025-01-06
feat: improve vertical sidebar with minification, theme mode and user menu when header is disabled

bastienrossi:
Hi,

I saw that an issue has been open for some time https://github.com/tabler/tabler/issues/1803 concerning the improvement of the sidebar in vertical mode.
I've made some modifications for my needs, so I'm proposing a PR with the code.
I've tried to use as little javascript as possible with the bootstrap collapse system, but I'd like your opinion on whether you think this is a good idea or whether it's preferable to use dedicated management.
I've added a sidebar.js script to manage tooltips when the sidebar is minimized.
I've also added a button to switch from dark to light mode and a user menu that opens over the sidebar to avoid conflicts.
I added these two elements because when the header is not in use, access to the user menu and dark mode is not possible.

Don't hesitate to give me feedback on what's wrong so that I can propose an improvement.

Here's how it looks:
menu-with-header-and-footer
menu-footer-opened
menu-minified
menu-minified-footer-opened
menu-minified-hover

Thanks


codecalm:
wow, it looks awesme!

cavasinf:
Would be a really great feature to have at the same time some `nav-header`, like `Projects` as @codecalm designed it:

image


They will not be visible when the navbar is collapsed:
image

codecalm:
Closing as no longer applicable — this touches `src/pages/_includes/layout/*.html`, a structure that predates even the `preview`/`docs` package split, let alone the later Eleventy → Astro migration (#2694). Feel free to reopen with a fresh PR against the current `preview/pages/*.astro` + `shared/astro/` structure if the sidebar improvement is still wanted.

PR #2054, opened 2025-01-11
Add theme settings

codecalm:
tabler-git-dev-props-tabler-io vercel app_(iPad Pro) (3)

tabler-git-dev-props-tabler-io vercel app_(iPad Pro)

tabler-git-dev-props-tabler-io vercel app_(iPad Pro) (1)

tabler-git-dev-props-tabler-io vercel app_(iPad Pro) (2)


carlosmintfan:
Amazing, this customizer! Now it's so easy to create a bit of an own branding while not modifying much of the SCSS of this template – or to let customers (children on LMS, for example) choose their own favourite color! :tada: 
Some feedback on improvements though: I'm planning to use the condensed layout, and, probably because of the increased spacing, it doesn't look that good: Paweł and Kuna in separate lines, 5844 over icons...navbar too big. If I change “5844 icons” to something shorter like "Icons" however, it gets better and "Paweł Kuna" is one line.

carlosmintfan:
And, on the Vercel preview of this PR I don't see tooltips on the world map anymore. This, however, happened to me even on the dev branch (not dev-props).

carlosmintfan:
Could we just remove the dark mode toggler from the navbar as it’s configurable in the settings offcanvas anyway? (Could fix the issue with the navbar in condensed layout, I‘ll test later.) (I think some people argue dark mode switches don’t belong on websites anyway as the user shall configure his preference in the operating system settings.)

codecalm:
@carlosmintfan  thank you for info! I'll update color variable as soon as possible 🙂

carlosmintfan:
@codecalm jsVectorMap isn't showing tooltips anymore. Google Chrome 132.0.6834.159 for Linux.
This is possibly even true for the dev branch when updating to the latest jsVectorMap, but can't promise it.

carlosmintfan:
@codecalm Would be great if you could fix this, the cards shouldn't be blue-ish if neither the primary nor the selected gray shade (stone) is. :)
Image demonstrating the blue-ish background color of cards in dark mode with orange primary color and stone shade

carlosmintfan:
@codecalm I'd recommend you to replace `text-secondary` by `text-body-secondary` about everywhere; it makes the contrast much better in dark theme with stone, for example.

carlosmintfan:
Also, `.hero-description`'s color should be `var(--tblr-secondary-color)`, not `var(--tblr-secondary)` for the same reason. Currently, the contrast is just bad in dark mode with stone theme base. This makes it much better.

carlosmintfan:
`link-secondary` also doesn't look good on dark yet, but there I'll need to dig a bit deeper.

kevinpapst:
This looks fantastic. How would you like to receive feedback @codecalm ? If everybody starts to drop 10 comments, you will be unable to follow our arguments... would you like to get review comments, so you can "resolve them"?

My biggest concern is more of a general feedback, I posted that in #2159 already:


> @codecalm I think it would be beneficial if you install a plugin like this and always double check changes: https://chromewebstore.google.com/detail/accessibility-insights-fo/pbjjkligggfmakdaogkfomddhfmpjeni
> 
> Being more or less compatible with WCAG would be a huge additional marketing argument. There are a lot of political regulations in place around the world, forcing companies to be WCAG compatible. Let's make it easier by shipping settings that respect WCAG by default. I'd be happy to support with testing!

martyncoup:
> This looks fantastic. How would you like to receive feedback @codecalm ? If everybody starts to drop 10 comments, you will be unable to follow our arguments... would you like to get review comments, so you can "resolve them"?
> 
> 
> 
> My biggest concern is more of a general feedback, I posted that in #2159 already:
> 
> 
> 
> 
> 
> > @codecalm I think it would be beneficial if you install a plugin like this and always double check changes: https://chromewebstore.google.com/detail/accessibility-insights-fo/pbjjkligggfmakdaogkfomddhfmpjeni
> 
> > 
> 
> > Being more or less compatible with WCAG would be a huge additional marketing argument. There are a lot of political regulations in place around the world, forcing companies to be WCAG compatible. Let's make it easier by shipping settings that respect WCAG by default. I'd be happy to support with testing!

+1 to this, I'd be happy to help as well.

codecalm:
@kevinpapst Papst this fature will be added to v1.2, so I don't finish it yet, but yes, you can add comments here, I'll resolve it when will be ready

core/scss/layout/_navbar.scss
kevinpapst, line 1: I can't find the exact source code location, so I share it here:
I would propose that you replace
```  --tblr-navbar-color: var(--tblr-secondary);```
with
```  --tblr-navbar-color: var(--tblr-body-color);```

Maybe its better to target `.navbar-nav {--tblr-nav-link-color}` instead which currently points to `var(--tblr-navbar-color)`. I am not sure which class is the correct one. 

This gives the navigation a much better contrast. This was discussed in #2159 before and multiple people (e.g. @mholt and @ethancrawford)

As there is the color border below, the active navigation is clearly distinguishable from the other nav items.

Before:
Bildschirmfoto 2025-03-01 um 16 16 53
Bildschirmfoto 2025-03-01 um 16 17 01

After:
Bildschirmfoto 2025-03-01 um 16 17 41
Bildschirmfoto 2025-03-01 um 16 17 19

carlosmintfan:
Selected color not respected for table background, and see how Comic looks on iOS:
Screenshot

codecalm:
@carlosmintfan I want to recreate Bootstrap styles, but in new way. I want to have similar classes, but independent of Bootstrap in future

codecalm:
Bootstrap has a lot of bugs, it's not actively developed. I can't use CSS variables because it uses old ways to modify color variables, etc. At this moment it's weakest part of development.

kevinpapst:
That's interesting insights, thanks for sharing. Probably the wrong place to discuss it further, so I just share this: BS 5.4 is almost there and development of BS6 has started 
https://github.com/orgs/twbs/discussions/41235#discussioncomment-12266169

carlosmintfan:
Thanks @codecalm, hopefully once the day will come when we have big MIT-licensed Tabler docs and don’t need the CC BY 3.0-licensed Bootstrap docs anymore. However, it seems you’re copying some short headers from Bootstrap docs to Tabler docs? Hmm, hopefully these are small enough to not fall under copyright.

PR #2077, opened 2025-01-24
Fix #2058

### LINKED ISSUE #2058: [BUG] Nav links are very very light now
### Browser

Firefox

### OS

Linux

### Screen size

Big

### Describe the bug

After beta 24, the nav links are very light:

image

And also are hard to see in dark mode:

image

I am not sure if that was intentional so I thought I'd open an issue about it. Thanks for Tabler!

### How to reproduce

(see demo site)

### Screenshots

_No response_

### JSFiddle

_No response_

farhan294sha:
Light mode
Screenshot 2025-01-24 081909

Dark mode

Screenshot 2025-01-24 081918

### DIFF
--- a/.changeset/fair-moose-enjoy.md
+++ b/.changeset/fair-moose-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Fix nav links color in dark mode
--- a/src/scss/_variables.scss
+++ b/src/scss/_variables.scss
@@ -53,9 +53,9 @@ $line-height-700: 3rem !default;
 
 $font-size-base: 0.875rem !default;
 
-$spacing-wide: 0.04em !default;
+$spacing-wide: .04em !default;
 $spacing-normal: 0 !default;
-$spacing-tight: -0.04em !default;
+$spacing-tight: -.04em !default;
 
 $body-letter-spacing: 0 !default;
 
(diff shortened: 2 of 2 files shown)


ethancrawford:
@farhan294sha @codecalm I might be wrong here but I suspect the solution in this PR is not quite relevant to the original bug...

My initial investigations led me to see that (one of?) the offending line(s?) that caused #2058 was this one:
https://github.com/tabler/tabler/blame/063ef5851531724d2c811dca20c355d8a77863f8/src/scss/_variables.scss#L758 - added while introducing the marketing page demo during https://github.com/tabler/tabler/commit/eb1b4f119f20f4413f30bd683b06af5f1af72c5b#diff-e5a2bc9e76b60200872fec420e32a890b2035706e458f23e9b901b4faad9ee6c.
I believe removing that line allowed the main demo site's navigation link colours to be restored as they were before the issue.
Doing so however also breaks the desired colour for the _marketing page's_ navigation links.
I'm not sure what the desired solution would be in this case, but my guess is that it is not to change the definition of `$text-secondary-*`? (Happy to be told otherwise! 🤷‍♂️ )

farhan294sha:
Oh, I haven’t looked at the marketing page. I thought it was a simple fix, and since this is my first contribution, I didn’t realize the full scope. You're right by removing [this line](https://github.com/tabler/tabler/blame/063ef5851531724d2c811dca20c355d8a77863f8/src/scss/_variables.scss#L758) fixes the home page, but the colors don’t match on the marketing page.

A suggestion might be to introduce another variable or adjust the `--#{$prefix}muted` value to ensure it has the correct contrast for both the home page and the marketing page navbar. If you have any suggestions or guidance on how to proceed, that would be really helpful!

carlosmintfan:
Does this really look that bad? (Lines https://github.com/tabler/tabler/blame/063ef5851531724d2c811dca20c355d8a77863f8/src/scss/_variables.scss#L758 and https://github.com/tabler/tabler/blame/063ef5851531724d2c811dca20c355d8a77863f8/src/scss/_variables.scss#L764 removed.)
Screenshot showing black navigation in marketing page
Isn't that a better compromise than the too-light links in the admin template?

carlosmintfan:
Hmm, but yeah, maybe it's also not the best practice if the navigation menu is brighter than the text between the heading and buttons.

codecalm:
I refactor navbar styles in #2159

PR #2078, opened 2025-01-24
Unknowncoderuse patch 2

unknowncoderuse:
Create jellky temple for better accessibility


codecalm:
Hi, unfortunately we don't use Jekyll to build sites anymore 🙁

PR #2085, opened 2025-01-27
Added 3 new payments from Nepal

mrthito:
This PR will add 3 new payment gateway from Nepal i.e. ESEWA, KHALTI and IMEPAY


ethancrawford:
@mrthito - looks like the existing list was sorted in alpha-numeric order - you might want to adjust the addition of your new list items to match?

codecalm:
@mrthito can you update SVGs to have the same size like other payments? (100x60px)

codecalm:
I've updated logos to match current styles. Thank you for your work! 🙂

PR #2108, opened 2025-01-29
Fixed the payment icon issue

mrthito:
In Reference to: https://github.com/tabler/tabler/pull/2085

There was a error passed through in my previous PR. So this will fix it


ethancrawford:
@mrthito - maybe I'm missing something, but I don't see the difference between this PR and the original one... what was the error you refer to?

PR #2109, opened 2025-01-29
Fixed the Payment JSON for payment code on khalti and esewa

mrthito:
In Reference to: https://github.com/tabler/tabler/pull/2085

There was a error passed through in my previous PR. So this will fix it


codecalm:
I've already fix it in 9a46a8fbf9ca0eecca94994dd276c3046f91448f

PR #2009, opened 2024-12-08
Replace .page-center with .my-auto in single page layouts

srsbiz:
Fixes #1820

PR #2114, opened 2025-01-30
Fix vertical alignment in single page and error layouts

### LINKED ISSUE #1820: [BUG] Terms of Service demo page is cut off

carlosmintfan:
The story:
- #2009 created to fix #1820 and merged. Vertical alignment is sacrificed.
- #2060 restores vertical alignment so is like a revert for #2009 except for using the `justify-content-center` class instead of `page-center`, but these are identical
- #2071 (yes, this one) removes the `justify-content-center` class from the single layout but not the error layout. (Now the single layout doesn't even have the `my-auto` class added in #2009.)

Current situation: error templates are vertically centered, single pages are not – the registration page just looks bad.
This PR fixes the issue: now the single (and error!) pages are correctly centered _if the screen is big enough_. How? Just works by removing `d-flex flex-column` classes from the body. Why? I'm not a Flexbox expert, but it just works. Not that I knew how to fix this, just randomly removed these classes from the body.


carlosmintfan:
@codecalm As this is trivial I‘d be happy if you just port this over into the files in the new monorepo structure, otherwise I‘ll update this later.

codecalm:
@carlosmintfan I've fixed, thank you 🙂

PR #2139, opened 2025-02-08
Refactor accordion component styles and markup, remove Bootstrap styles

codecalm:


carlosmintfan:
@codecalm The jsVectorMap does not show tooltips anymore on the Vercel preview. It possibly even doesn't work when updating to the latest version of it on the dev branch, but I can't promise it. Using Google Chrome 132.0.6834.159 on Linux.

PR #2141, opened 2025-02-09
Refactor alert component styles and markup, remove Bootstrap styles

codecalm:


kevinpapst:
I missed the PR before, so this feedback is a bit late...
Thanks for working on the revamped alerts @codecalm  and sorry for not having a look earlier.

**Good**
The readability of the normal versions got much better, because there are no colors used anymore for the fonts 👍 
Especially dark mode is so much better now!

And I like that the playful bits were removed, looks cleaner now. I assume the colored borders could still be added with some classes.

**What I would like to change**
The "important" version is not popping out any more, I use them for toast messages and they will not work like that.
But even more important: they are hard to read now. Regular mode is already off, but dark mode is 100% worse. I cannot read the new danger+warning at all. I am wearing glasses, and the red and orange font is completely blurry.

I would vote to revert the changes for the important alerts.
Not sure if that is only a vision problem for me, but I
If you prefer the new version, maybe you can add them as additional classes?

**Before**
Bildschirmfoto 2025-02-28 um 00 36 59

Bildschirmfoto 2025-02-28 um 00 39 27

**After**
Bildschirmfoto 2025-02-28 um 00 37 11

Bildschirmfoto 2025-02-28 um 00 39 16

PR #2159, opened 2025-02-13
Enhance navbar styles with new hover effects and color variables

codecalm:
Fixes: #2058

Screenshot 2025-02-13 at 23 46 35
Screenshot 2025-02-13 at 23 46 28

### DIFF
--- a/.changeset/thick-kids-glow.md
+++ b/.changeset/thick-kids-glow.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Fixes navbar styles with new hover effects and color variables
--- a/core/scss/_variables.scss
+++ b/core/scss/_variables.scss
@@ -754,7 +754,7 @@ $nav-tabs-bg: var(--#{$prefix}bg-surface-tertiary) !default;
 // Navbar
 $navbar-height: 3.5rem !default;
 $navbar-padding-y: 0.25rem !default;
-$navbar-light-color: var(--#{$prefix}muted) !default;
+$navbar-light-color: var(--#{$prefix}secondary) !default;
 
 $navbar-hover-color: $white !default;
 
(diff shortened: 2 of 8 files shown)


ethancrawford:
Just to clarify, @codecalm - was the idea behind the variable change from 'muted' to 'secondary' purely a naming thing, to basically assign the nav link color variable a more appropriate label?

@mholt - what do you think of the adjusted colour scheme? I definitely appreciate the nav links in dark mode being slightly easier to see.

Maybe it's just me, but I can't help but wonder whether navbar links, active or not, need slightly more contrast still? (although the _active_ nav item stands out well from the inactive ones, the light mode links are still the same colour as they were when the issue was raised, which I would have thought would have benefitted from being a bit higher contrast). 😅 What do you folks think?

ethancrawford:
Looks like Tabler Props will make things easier to maintain for sure 🙂

mholt:
@ethancrawford Looks better; I still think the contrast is a _bit low_ on those links, but it could be that my eyes are starting to go bad already. I dunno. I wouldn't mind an extra nit of contrast personally. Still an improvement though!

Thanks @codecalm

kevinpapst:
@mholt you are not alone here. The navigation is a central place and should be as clear as possible.
In dark mode white links would be the best, the active menu is highlighted with the colored underline.

Also this is a huge accessibility issue, not everyone is 24 and has perfect eyesight. I have a hard time reading these links.

@codecalm I think it would be beneficial if you install a plugin like this and always double check changes:
https://chromewebstore.google.com/detail/accessibility-insights-fo/pbjjkligggfmakdaogkfomddhfmpjeni

Being more or less compatible with WCAG would be a huge additional marketing argument. 
There are a lot of political regulations in place around the world, forcing companies to be WCAG compatible. Let's make it easier by shipping settings that respect WCAG by default. I'd be happy to support with testing! 

Right now the plugin says the same, the contrast is too low on all the grey texts in dark mode (not only navigation, but this part ic crucial):
Bildschirmfoto 2025-03-01 um 08 11 45

PR #2208, opened 2025-03-10
Refactor autosize and input mask plugins to use window scope

codecalm:


QThans:
When can this pr be released in an new version? Waiting for it.

Issue #2211, opened 2025-03-12
[BUG] Badge in navbar text-secondary


vjmartins:
### Browser

All

### OS

All

### Screen size

All

### Describe the bug

Badge in navbar text-secondary

### How to reproduce

Open page https://preview.tabler.io/navigation.html

### Screenshots

![Image](https://github.com/user-attachments/assets/bab3c019-8fe8-4cf7-b2b1-05ca605b7213)

ethancrawford:
@vjmartins - I may have misunderstood, and I certainly don't speak for the project maintainers - but it seems to me that the badge has been on this menu item for a long time, and was introduced as a deliberate feature in https://github.com/tabler/tabler/commit/1b0f32833b2ec86a7354ddd58a6fbf712071091d. (Perhaps just to demonstrate the use of a badge on a menu item).
In what sense do you feel it's a bug?

vjmartins:
I think colour should be white.

ethancrawford:
Given that the badge label colour was updated to `--tblr-secondary` as recently updated here:
https://github.com/tabler/tabler/commit/378fba89f58d56bd12b9dd7d2efd3ff71c5dbfcf#diff-26daa114aa235a9e78fa23013fab7be95cbb62977b6d2ebf4c6eeca20c093b0cR566
Possibly the intention is to use something more like `secondary-fg` instead of `secondary` on this line:
https://github.com/tabler/tabler/blob/922bb0346a45d9bf5d6d708dc0c4a66f25c68d33/core/scss/_variables.scss#L566
...which in turn currently resolves to `--tblr-light`.
Whether _that_ is the correct colour and/or solution I guess will need confirmation from the maintainers 🤷‍♂️

krish144:
I'm a beginner and eager to contribute. I changed the color in the _variable.scss file. Can I create a pull request again? Also, could you please share any tips for contributing effectively and collaborating with other contributors?

PR #2216, opened 2025-03-15
Change badge text color in navbar

krish144:
Before:
Tabler-Premium-and-Open-Source-dashboard-template-with-responsive-and-high-quality-UI--03-15-2025_05_16_AM

After:
Tabler-Premium-and-Open-Source-dashboard-template-with-responsive-and-high-quality-UI--03-15-2025_05_14_AM (1)

### DIFF
--- a/core/scss/_variables.scss
+++ b/core/scss/_variables.scss
@@ -563,7 +563,7 @@ $badge-font-weight: var(--#{$prefix}font-weight-medium) !default;
 $badge-padding-y: 0.25em !default;
 $badge-padding-x: 0.5em !default;
 $badge-empty-size: 10px !default;
-$badge-color: var(--#{$prefix}secondary) !default;
+$badge-color: var(--#{$prefix}light) !default;
 $badge-bg-color: var(--#{$prefix}bg-surface-secondary) !default;
 
 // Buttons
--- a/preview/pages/_includes/layout/navbar-menu.html
+++ b/preview/pages/_includes/layout/navbar-menu.html
@@ -25,7 +25,7 @@
 				</span>
 
 				{% if level-1[1].badge %}
-				<span class="badge badge-sm bg-red">{{ level-1[1].badge }}</span>
+				<span class="badge badge-sm bg-red text-light">{{ level-1[1].badge }}</span>
 				{% endif %}
 			</a>
 
--- a/preview/pages/_includes/layout/navbar-menu.html
+++ b/preview/pages/_includes/layout/navbar-menu.html
@@ -25,7 +25,7 @@
 				</span>
 
 				{% if level-1[1].badge %}
-				<span class="badge badge-sm bg-red text-light">{{ level-1[1].badge }}</span>
+				<span class="badge badge-sm bg-red">{{ level-1[1].badge }}</span>
 				{% endif %}
 			</a>
 


ethancrawford:
@krish144 - I can't speak for the project maintainers of course, but while this does change the colour to something more legible, I'm not convinced this particular solution is the right one for the issue. As described in https://github.com/tabler/tabler/issues/2211, the basic problem is that the badge label is taking its colour from `--tblr-secondary`, as updated recently here:
https://github.com/tabler/tabler/commit/378fba89f58d56bd12b9dd7d2efd3ff71c5dbfcf#diff-26daa114aa235a9e78fa23013fab7be95cbb62977b6d2ebf4c6eeca20c093b0cR566
... and therefore, possibly, `--tblr-secondary` is not always the right color to use for the badge label.
So, I suspect it's probably a deeper issue than just needing to override the badge colour in the demo's html - as badges used elsewhere would still have the same issue.

As to what the correct solution actually is, I'm not entirely sure, but I've added my thoughts on #2211 🙂

krish144:
> @krish144 - I can't speak for the project maintainers of course, but while this does change the colour to something more legible, I'm not convinced this particular solution is the right one for the issue. As described in #2211, the basic problem is that the badge label is taking its colour from `--tblr-secondary`, as updated recently here: [378fba8#diff-26daa114aa235a9e78fa23013fab7be95cbb62977b6d2ebf4c6eeca20c093b0cR566](https://github.com/tabler/tabler/commit/378fba89f58d56bd12b9dd7d2efd3ff71c5dbfcf#diff-26daa114aa235a9e78fa23013fab7be95cbb62977b6d2ebf4c6eeca20c093b0cR566) ... and therefore, possibly, `--tblr-secondary` is not always the right colour to use for the badge label. So, I suspect it's probably a deeper issue than just needing to override the badge colour in the demo's html - as badges used elsewhere would still have the same issue.
> 
> As to what the correct solution actually is, I'm not entirely sure, but I've added my thoughts on #2211 🙂


First of all, I apologize for any confusion caused, and I sincerely appreciate your advice. I’m still relatively new to contributing, and I initially thought that pushing the code would be enough to resolve the issue. However, I’ve come to realize that I made an incorrect assumption. Thank you again for your guidance, and I’ll make sure to follow the correct steps moving forward.

ethancrawford:
@krish144 - you didn't do anything bad - all good! Sometimes the right solution isn't clear, but (often anyway 😆 ) the more experience you gain, the easier that becomes 🙂

ethancrawford:
@krish144 - sorry if it wasn't clear in my comments on #2211. I suggested there that I thought the solution might be to replace `secondary` on that line with `secondary-fg`. (Part of my thinking was that It's possible, though can't be sure, that this was the original intention by @codecalm but was a simple typo).

My guess also is that we probably wouldn't use `--tblr-light` here because Tabler is designed for light and dark themes, (and possibly more in future??) so it's maybe not guaranteed that the text will always be 'light'. Using `secondary-fg` on the other hand fits in better with multiple themes, since it in turn could refer to another colour that is light _or_ dark.

(In the end, only codecalm or possibly the other maintainers really know exactly what colour is desired, so we can either make a best guess that using `secondary-fg` is correct, use it, and accept that they may want to use something else, or - we can ask the maintainers for feedback first. It's just a judgement call between how much you think the suggested solution is correct, and how much effort is involved in possibly redoing it if it isn't 🤷)

codecalm:
I've fixed it in #2236

PR #2227, opened 2025-03-19
Fix broken shape in kr.svg

YamiOdymel:
While reviewing country flags for #2226, I noticed that the South Korean flag was correct until https://github.com/tabler/tabler/commit/00446acebb9d1f3afd7023c5723ba1a5939f89bd broke it, causing it to resemble a weird shield 🛡️ icon.

This pull request restores the original Taegeuk symbol while keeping the colors and "bars" unchanged from the current version.

---

**Current (Broken)**

image

---

**Old**

image

---

**New**

image


codecalm:
thank you so much for help! 🙂

Issue #2225, opened 2025-03-19
[BUG] Missing RTL CSS File in Theme Layout


MohammedFuad97:
### Browser

Chrome

### OS

Windows 11

### Screen size

not relative

### Describe the bug

I am facing an issue with the RTL layout in the theme. When I open the page: /preview/dist/layout-rtl.html

The following CSS file is missing, resulting in broken styles:

Console error: /preview/dist/preview/css/demo.rtl.min.css?1740918423 net::ERR_FILE_NOT_FOUND

### How to reproduce

1. Go to '/preview/dist/layout-rtl.html'
2. Open inspect element
3. You'll find the missing asset

### Screenshots

![Image](https://github.com/user-attachments/assets/e5a22151-9a24-4016-b4dd-a969f98a322b)

PR #2231, opened 2025-03-20
Delete missing demo rtl style

mikosamuraj:


ethancrawford:
Granted this might remove the related error, but if there are indeed 'broken styles' as referred to in #2225, hopefully they can be restored/fixed at some point... 🤞

PR #2238, opened 2025-03-24
Refactor disabled color variables in navbars

codecalm:


kevinpapst:
Much better now, thanks 👍

PR #2239, opened 2025-03-24
Refactor SCSS for alerts and close button styles

codecalm:


kevinpapst:
Thanks for taking a look at this! Much better now. Just one conern:
 
I can still not read the important danger and warning alerts in dark mode, the contrast of the black text  for these two is very bad:
image

A second minor issue: the background color of non-important warning and danger is very alike, I can hardly keep them apart

codecalm:
@kevinpapst fixed:

Screenshot 2025-03-25 at 23 03 38
Screenshot 2025-03-25 at 23 03 32

PR #2274, opened 2025-04-19
Update icons to v3.31.0

codecalm:


ethancrawford:
@codecalm - it would be great if we can merge/release this one soonish as I suspect it would potentially help solve one or two open issues - will you have a chance soon? :pray:

PR #2277, opened 2025-04-20
fix: add missing border-radius to .card-header-tabs

GuillermoFarias:
## What this PR does

This PR adds a missing `border-radius` to the `.card-header-tabs` class to match the style applied to `.card-header:first-child`. This ensures visual consistency across card headers that use tabs.

## Why it's needed

Without this fix, `.card-header-tabs` has sharp corners, which breaks the rounded card appearance and leads to inconsistent UI.

### Before
image

### After
image

## Notes
Let me know if this should be applied in a different place or if there's a related variable that should be used instead.
First-time contributor — thanks for the awesome work on Tabler!

### DIFF
--- a/.changeset/tidy-hairs-wonder.md
+++ b/.changeset/tidy-hairs-wonder.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Fix missing border-radius to `.card-header-tabs`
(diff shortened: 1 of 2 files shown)


ethancrawford:
@GuillermoFarias - out of curiosity, had you already seen the example card with header tabs at https://preview.tabler.io/cards.html ? (if so, does that one not suit your use case?)

GuillermoFarias:
@ethancrawford 
> @GuillermoFarias - out of curiosity, had you already seen the example card with header tabs at https://preview.tabler.io/cards.html ? (if so, does that one not suit your use case?)

Yes, I had seen that example, but it shows the same behavior.

image

This simple fix solves the border-radius issue consistently in all cases.

codecalm:
thank you for help!

PR #2280, opened 2025-04-23
Fix transparent background on alerts after v1.2.0 update

anischelly26:
I noticed that after updating to v1.2.0, the alert backgrounds were turning transparent, especially when changing the card color — it didn’t look great visually.

I’ve added a small fix by adjusting the CSS variables for the alert background and border using color-mix. This brings back proper visibility for all alert types.

Tested it locally and it seems to do the trick! Let me know if any adjustments are needed

### DIFF
--- a/core/scss/_bootstrap-override.scss
+++ b/core/scss/_bootstrap-override.scss
@@ -76,3 +76,7 @@
 @function opaque($background, $foreground) {
   @return mix(rgba($foreground, 1), $background, opacity($foreground) * 100%);
 }
+.alert {
+  --tblr-alert-bg: color-mix(in srgb, var(--tblr-alert-color) 20%, var(--tblr-bg-surface));
+  --tblr-alert-border-color: color-mix(in srgb, var(--tblr-alert-color) 40%, var(--tblr-bg-surface));
+}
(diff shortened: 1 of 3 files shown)


ethancrawford:
@anischelly26 - the project maintainers may have further advice, but the changes to [package.json](https://github.com/tabler/tabler/pull/2280/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) and [pnpm-lock.yaml](https://github.com/tabler/tabler/pull/2280/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb) in this PR are unrelated to this specific issue, and should probably be removed.
Also, I don't know for sure, but it would seem to me that [core/scss/_bootstrap-override.scss](https://github.com/tabler/tabler/pull/2280/files#diff-ffc1e4cc45bb4a405310e34802966901699b2a1775d27fe1f6fc07def479d622) is not the best place to add the CSS variable change - possibly [core/scss/ui/_alerts.scss](https://github.com/tabler/tabler/blob/dev/core/scss/ui/_alerts.scss)? (Though again, I'm not 100% sure)...

PR #2281, opened 2025-04-23
Fix WYSIWYG dark mode detection

anischelly26:
Hey folks ,I stumbled upon the WYSIWYG dark mode issue and saw that the problem comes from checking localStorage for tablerTheme instead of tabler-theme.

Rather than just swapping the key, I figured it’d be safer to handle both cases — just in case someone out there still has the old key saved. So now, it checks for both. Tested it locally, and dark mode kicks in as expected


ethancrawford:
@anischelly26 - did you mean to include a different git commit here? the commit in this PR is the same one as in https://github.com/tabler/tabler/pull/2280

PR #2306, opened 2025-04-30
Update icons version and import script data path

### LINKED ISSUE #2297: [BUG] Switch icon examples are broken
### Browser

N/A

### OS

N/A

### Screen size

N/A

### Describe the bug

When the examples on the switch icons documentation page are clicked, they disappear. Instead, the second icon in the switch should appear.

I notice in the html of the examples, that the `switch-icon-a` elements have svg icons nested in them, but the `switch-icon-b` elements are empty.

Should the `switch-icon-b` elements have svg icons nested in them instead?

### How to reproduce

1. Go to https://docs.tabler.io/ui/components/switch-icon
2. Click on any of the switch icon examples
3. Observe that the switch icons disappear instead of being replaced by `switch-icon-b` icons.

### Screenshots

_No response_

### JSFiddle

_No response_

ethancrawford:
I'm not totally familiar with the process of importing the latest icons into Tabler core/preview, so @codecalm / @BG-Software-BG or other folks more familiar with the process, feel free to comment on anything that's not right - but I decided to test whether I could manually update https://github.com/tabler/tabler/blob/dev/shared/data/icons.json by running `preview/.build/import-icons.mjs` on the command line.

At first, I encountered an error because the path here did not exist:
https://github.com/tabler/tabler/blob/e6753898714306509e5cdb4e566897b387be3e12/preview/.build/import-icons.mjs#L29
(which in my case was returning):
`Error: ENOENT: no such file or directory, open '/home/ethan/repos/tabler/preview/shared/data/icons-info.json'`
instead of generating the correct directory, `/home/ethan/repos/tabler/shared/data/icons-info.json`.

Once I updated the path on lines 29 and 38 of `/preview/.build/import-icons.mjs`, the `/shared/data/icons.json` and `/shared/data/icons-info.json` files appeared to update successfully.

I was intending to test whether updating `/shared/data/icons.json` would fix #2297 - however I'm also not sure how to view a local dev version of the documentation app - is this currently possible?


ethancrawford:
😂 😂 
That's what I get for not checking the existing open PR list. Oops. Sorry!

PR #2318, opened 2025-05-02
Fix input mask example in docs

### LINKED ISSUE #2308: [BUG] Import mask documentation page example broken
### Browser

N/A

### OS

N/A

### Screen size

N/A

### Describe the bug

The live example on the [import mask documentation page](https://docs.tabler.io/ui/forms/form-input-mask) does not work.

This was possibly meant to be updated as part of #2208.

I'm not 100% sure, but it's possible the fix is just adding a line saying:
```
docs-libs: imask
```
...into the front matter section of https://github.com/tabler/tabler/blob/dev/docs/content/ui/forms/form-input-mask.md.

(I'd do so myself and test to make sure, but it's not clear how to test a local version of the documentation app).

### How to reproduce

1. Go to https://docs.tabler.io/ui/forms/form-input-mask#default-markup
2. Scroll down to the live demo (with the label 'Telephone mask')
3. Type anything into the text box, such as `aaaaa`
4. Observe that the input is not restricted

### Screenshots

Image

### JSFiddle

_No response_

ethancrawford:
There was a missing lib that needed to be added to the front-matter of the form input mask documentation page.

Fixes #2308

### DIFF
--- a/.changeset/dry-spiders-double.md
+++ b/.changeset/dry-spiders-double.md
@@ -1,0 +1,5 @@
+---
+"@tabler/docs": patch
+---
+
+Fix input mask example in docs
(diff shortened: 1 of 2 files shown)


ethancrawford:
Input mask text box before this fix:
image

Input mask text box after this fix:
image

(typing something like `aaaaaa` in the box is now prevented).

PR #2321, opened 2025-05-03
Fix colour picker preview page not displaying correctly

mikosamuraj:
Screenshot 2025-05-03 at 22 45 48

The problem was caused by fetching by production app lib `melloware` instead of `@melloware` - `shared/includes/layout/js-libs.html` is doing replace `@` for '' . In case of not breaking other libs i registered mapping via `addPassthroughCopy`.


kevinpapst:
Can confirm that https://preview.tabler.io/colorpicker.html is currently broken, because the URL https://preview.tabler.io/libs/melloware/coloris/dist/umd/coloris.min.js?1745688879 leads to a 404

kevinpapst (commented):
Wouldn't it be better to remove the @ from the target folder here: 
https://github.com/tabler/tabler/blob/dev/shared/e11ty/data.mjs#L33C10-L33C37
I don't know the code, so its just a wild guess, but @melloware is the only one with a @ right now: https://github.com/tabler/tabler/blob/dev/shared/data/libs.json

melloware:
Happy to hear someone is using the library!

PR #2330, opened 2025-05-05
Fix carousel with captions demo on preview website

ethancrawford:
It seems that previously a prop was not being passed to the relevant carousel include file to enable captions for that specific example.


ethancrawford:
Example before this change:
image

Example after this change:
image

kevinpapst (approved):
Seems this is only used in one place and a bool is passed:
https://github.com/tabler/tabler/blob/e6753898714306509e5cdb4e566897b387be3e12/preview/pages/carousel.html#L20

Which is evaluated here:
https://github.com/tabler/tabler/blob/dev/shared/includes/ui/carousel.html#L19

LGTM 👍

PR #2333, opened 2025-05-05
Update heading of carousel with thumbnail indicators

ethancrawford:
The carousel example that shows vertical thumbnail indicators had the same heading as the example with horizontal thumbnail indicators. This change updates the vertical thumbnail example's heading.
(This follows the pattern of the dot indicator examples, which have different headings).

Before this change:
image

After this change:
image


kevinpapst:
Thanks @ethancrawford  👍

PR #1492, opened 2023-02-27
New docs

codecalm:
preview: https://tabler.io/docs/

Screenshot 2023-02-27 at 01 09 04

PR #1391, opened 2022-11-29
add input with appended `<kbd>`

codecalm:
Screenshot 2022-11-29 at 22 40 33

PR #2334, opened 2025-05-05
Fix incorrect label text on form elements docs page

ethancrawford:
The example of an input with an appended `<kbd>` previously had a change introduced in https://github.com/tabler/tabler/pull/1492 to its label that was incorrect. 

This commit brings it back to what it was originally, as first written in https://github.com/tabler/tabler/pull/1391.

Before this change:
image

After this change:
image

### DIFF
--- a/.changeset/short-balloons-smile.md
+++ b/.changeset/short-balloons-smile.md
@@ -1,0 +1,5 @@
+---
+"@tabler/docs": patch
+---
+
+Fix incorrect label text on form elements docs page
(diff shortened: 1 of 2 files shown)


kevinpapst:
I have no clue: is this still used?
https://github.com/tabler/tabler/blob/dev/docs/content/ui/forms/form-elements.md?plain=1#L554

ethancrawford:
There is a corresponding example on the preview site at https://preview.tabler.io/form-elements.html:
image

kevinpapst:
Sorry, what I actually meant: could you update the other doc as well, because it also looks wrong

kevinpapst:
Excuse me, I got confused between several branches. Ignore my question!

PR #2345, opened 2025-05-07
Improve card footer layout and enhance entry display format in invoices

mikosamuraj:
Screenshot 2025-05-08 at 00 10 28


ethancrawford:
@mikosamuraj - how does the change look on larger screens? Any chance of another screenshot?
(Just wondering what the behaviour should be on non-mobile screens, since we'd probably have more room to keep the 'page X of Y' on the same horizontal alignment as the pagination controls at that size)

PR #2347, opened 2025-05-08
Use color-input include file more

ethancrawford:
Previously, an include file was created to generate a collection of coloured input controls. This was used in several places, but there were other places where nearly identical input collections were still defined without using the include file. This change uses the include file in more places.

The labels on the colour inputs on the preview's live form demo have also been updated to be a little clearer.


ethancrawford:
This also fixes an issue on the documentation's form color check page where the first example is intended to act like a collection of _checkboxes_, but instead acts like a collection of _radio buttons_.

PR #2350, opened 2025-05-09
Fix colour swatches on small screens

### LINKED ISSUE #2327: [BUG] Colour swatches in colours docs overlap each other on small screen
### Browser

N/A

### OS

N/A

### Screen size

360 x 800 (for example)

### Describe the bug

On small screen devices, the colour swatches on the colours documentation page overlap each other.
This is due to a fixed padding applied to each swatch both on larger and smaller screens, set here:
https://github.com/tabler/tabler/blob/e6753898714306509e5cdb4e566897b387be3e12/shared/includes/docs/colors.html#L4

Perhaps the level of padding should be responsive.

### How to reproduce

1. Using a mobile device, or desktop browser with screen set to emulate a mobile device of say, 360 x 800 pixels, visit https://docs.tabler.io/ui/base/colors
2. Scroll down to the colour examples
3. Observe that the colour swatches overlap each other.

### Screenshots

Image

### JSFiddle

_No response_

ethancrawford:
Previously, a fixed padding amount was causing the colour swatches to overlap each other on small screens. They are now responsive: on small screens, they shrink slightly and spread out more across each line of their layout.

The markup was also changed slightly to make use of the `row-cols-*` shortcuts.

Fixes #2327.

### DIFF
--- a/.changeset/poor-spoons-deny.md
+++ b/.changeset/poor-spoons-deny.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": "patch"
+---
+
+Fix colour swatches on small screens
--- a/shared/includes/docs/colors.html
+++ b/shared/includes/docs/colors.html
@@ -1,7 +1,7 @@
-<div class="row g-4">
+<div class="row row-cols-4 row-cols-md-6 g-3 g-md-4">
 	{%- for color in include.colors -%}
-	<div class="col-2 text-center">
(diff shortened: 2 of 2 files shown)


ethancrawford:
Swatches on small (eg mobile) screens before this change: See #2327.

After this change:
Screenshot 2025-05-09 111541

PR #2363, opened 2025-05-11
Add FullCalendar integration with sample events and styling

codecalm:
localhost_3000_fullcalendar html
localhost_3000_fullcalendar html_theme=dark


preview/package.json
```diff
@@ -42,6 +42,7 @@
     "dropzone": "^6.0.0-beta.2",
     "flatpickr": "^4.6.13",
     "fslightbox": "^3.5.1",
+    "fullcalendar": "^6.1.17",
```
kevinpapst, line 45: not sure about the difference, but I use `@fullcalendar`

BG-Software-BG:
It would be a good idea to include a month/week/day switch. 

I agree that it should look better in the mobile view. I think that `contentHeight: 'auto'` is a better approach than the current sized calendar (with a scrollbar). Maybe the day view should be the default one on mobile?

I don't see any more issues.

codecalm:
I’ll create next PR with this upgrades, because I had problems on mobile devices and I need more time to do it, but I want to release next version today

vjmartins:
I suggest wait to the v7 launch - https://github.com/fullcalendar/fullcalendar/issues/7907

PR #2366, opened 2025-05-12
feat: Add new charts to dashboard pages

codecalm:
Screenshot 2025-05-12 at 12 42 52


shared/includes/layout/homepage.html
```diff
@@ -1,4 +1,36 @@
...
+		<div class="card">
+			<div class="card-body">
+				<div class="subheader">Total Users</div>
+				<div class="d-flex align-items-baseline">
+					<div class="h1 mb-0 me-2">75,782</div>
+					<div class="me-auto">
+						{% include "ui/trending.html" value=2 %}
+					</div>
+				</div>
+				<div class="text-secondary mt-2">24,635 users increased from last month</div>
```
ethancrawford, line 15: Had a quick look by building this locally, looks good to me.

I might have one optional suggestion. Maybe it's being super picky though, I'm happy to hear your feedback!
When I see two cards both labelled with "Active Users", where one says "25,782" and the other says "2,986", it's not clear if those two numbers represent two different metrics (like one being all-time for example, the other just in 7 days), or just a typo of some sort. (It's not, and I get that this is just a demo site, but if I was being _really_ picky I'd suggest making it clearer that these were two different metrics, so that it didn't potentially look confusing 🙂) thoughts?

ethancrawford (approved):

shared/includes/layout/homepage.html
kevinpapst: This is indeed on the extreme side of picky 😁

ethancrawford: :joy: happy either way 😉

BG-Software-BG: I never looked at all those titles and numbers but... changed to "Active subscriptions" as @kevinpapst suggested.

BG-Software-BG:
Mobile view fixed

image

codecalm:
everything fixed!

PR #2371, opened 2025-05-13
fix(layout): maintain sticky navbar by removing conflicting height declaration

### LINKED ISSUE #2332: [BUG] Sticky nav on documentation site only partially works
### Browser

N/A

### OS

N/A

### Screen size

2560 x 1440

### Describe the bug

If you view a page in the documentation where the content takes up more than two screen's-worth of height, then the sticky top navigation only stays sticky until you scroll past the first screen's worth of content.
I think this is because of this line here:
https://github.com/tabler/tabler/blob/b6ce7b8feb897a832ca3323ee08921a4ca8c05e4/core/scss/layout/_core.scss#L9

Removing it allows the sticky nav to remain visible past one screen's-worth of content, though more investigation may be needed to determine if removing the above line also has other side effects.

### How to reproduce

1. View a page on the documentation site that is more than twice as large as your current screen resolution height. For my resolution height of 1440, https://docs.tabler.io/ui/base/typography is more than enough.
2. Scroll down past one entire screen's-worth of content. On my screen, this is about half way through the live demo under the heading "Paragraphs".
3. Observe that the sticky top nav bar scrolls up and out of view.

### Screenshots

Image

### JSFiddle

_No response_

anischelly26:
Hey team!  

This fixes #2332 where the navbar would stop being sticky after scrolling past one screen height. Turns out the `height: 100%` in `_core.scss` was accidentally creating a new scroll container, which broke the sticky behavior.  

The fix is simple:  
- Removed the `height: 100%` line from `_core.scss`  
- Let the navbar stick naturally to the viewport  

I tested this by:  
- Scrolling through long pages in the docs (like the Typography page)  
- Checking mobile + desktop layouts  
- Making sure other sticky elements (like sidebars) still work  

Before:  
Navbar vanished when scrolling down 
After:
Navbar stays put like a champ!


ethancrawford:
@anischelly26, this PR contains changes that are not related to #2332.
As Kevin mentioned a few weeks ago, you should read about Git and Pull requests before opening further PRs, so that you only provide changes that are specific to the task, instead of a whole pile of changes that are related to many other things at the same time.

PR #2384, opened 2025-05-17
Enhance pagination component with new styles and documentation

codecalm:


docs/content/ui/components/pagination.md
```diff
@@ -0,0 +1,60 @@
...
+If you want to use outline version of pagination, you can use `.pagination-outline` class. This will give you a more subtle look and feel, which is great for applications where you want to keep the focus on the content rather than the navigation.
+
+{% capture html -%}
+{% include "ui/pagination.html" class="pagination-outline" %}
+{%- endcapture %}
+{% include "docs/example.html" html=html centered vertical %}
+
+## Circle version
+
+If you want to use circle version of pagination, you can use `.pagination-circle` class. This will give you a more subtle look and feel, which is great for applications where you want to keep the focus on the content rather than the navigation.
```
ethancrawford, line 53: I'd suggest adding a note about outline version to this, and then swap the order of the two circle examples below it.
```suggestion
If you want to use a circle version of the pagination, you can use the `.pagination-circle` class. This will give you a more subtle look and feel, which is great for applications where you want to keep the focus on the content rather than the navigation. This can also be combined with the `.pagination-outline` class for a more prominent look.
```

docs/content/ui/components/pagination.md
```diff
@@ -0,0 +1,60 @@
...
+
+{% capture html -%}
+{% include "ui/pagination.html" class="pagination-circle pagination-outline" -%}
```
ethancrawford, line 56: As mentioned above, I'd suggest swapping this line and line 58 (the 'plain' circle example). That way you're showing the basic circle first, and then adding on the outline to it.

codecalm:
@ethancrawford I've added your changes to `dev` branch

PR #2385, opened 2025-05-17
Move additional libraries to Tabler core

codecalm:


kevinpapst:
What are the consequences of that change? 
I don't understand all the internal build files: does that change the released file structure?

codecalm:
yes, from now all librares will be served from `@tabler/core` instead of outside libraries.

kevinpapst:
Let me rephrase the question: will there still be vendor files in the Tabler npm package?
I am just wondering about BC breaks that should be documented

codecalm:
yes, I don't know is is a breaking change, because old links will be works still. This is mainly for the comfort of the users, so that they could be sure always that the vendor lib version was definitely compatible with Tabler

kevinpapst:
Aha, so you are inlining the vendor files?
But there will still be vendors.css / js file?

codecalm:
yes, yes

from now you can include library from eg. `@tabler/core/dist/libs/apexcharts/apxcharts.min.js` instead of `apexcharts/apxcharts.min.js`

If you upgrade Tabler all vendor libraries will update automatically

PR #2388, opened 2025-05-18
Fix FAQ accordion structure

codecalm:


kevinpapst (approved):
Yes, now it looks good again 👍

PR #2397, opened 2025-05-21
List group item hoverable only with .list-group-hoverable class

vitek-dev:
In "Last commits" example (https://preview.tabler.io/lists.html) there is example with classes "list-group list-group-flush list-group-hoverable" but the fact is that "list-group-hoverable" is ignored and items are hoverable even without that.


codecalm:
Thank you for your help! 🙂

PR #2402, opened 2025-05-22
Update class names from left, right to start, end

codecalm:
Replace all classes ended by `*-left`, `*-right` to `*-start`, `*-end` with fallback

### DIFF
--- a/.changeset/blue-spoons-sleep.md
+++ b/.changeset/blue-spoons-sleep.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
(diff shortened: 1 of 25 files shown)


ethancrawford:
@codecalm I think there are still references to an `icon-right` class that need updating too.

kevinpapst:
Thanks for keeping the "wrong" ones (left,right) until the next major version.

codecalm:
@kevinpapst @ethancrawford I've update classnames in all files, can you check is okay now?

PR #2413, opened 2025-05-28
wysiwyg.html: Remove license_key option from HugeRTE init object

carlosmintfan:
This is a leftover of TinyMCE 7 initialization; HugeRTE is provided under the MIT license.


codecalm:
thank you for your help! ❤️

PR #2430, opened 2025-06-10
Rename several Theme builder labels

ethancrawford:
We have updated the heading on the top nav button, the heading inside the theme
settings side panel, and the side panel's save button.

There were a couple of reasons why this change was recommended:
1)  the name 'Settings' in the top nav was not really clear about what _kind_ of
settings it represented
2) There is already a link in the 'Extra' nav menu called 'Settings', and having
both items named the same was potentially confusing.
3) Theme Settings seemed less jargon-y and more straightforward for
non-technical folks to understand than Theme Builder.
4) 'Save' seemed shorter and still simple enough to understand over 'Save
settings'.
5) The settings icon on the save button has been removed as it seemed
superfluous and/or unrelated.

### DIFF
--- a/shared/includes/layout/navbar.html
+++ b/shared/includes/layout/navbar.html
@@ -64,7 +64,7 @@
 											{% include "ui/icon.html" icon="settings" %}
 										</span>
 										<span class="nav-link-title">
-											Settings
+											Theme Settings
 										</span>
 									</a>
 								</li>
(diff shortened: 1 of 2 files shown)


kevinpapst:
Well, if we’re being picky: the button should probably just be `Save`.
And the cogs icon doesn’t make sense — a Save action doesn’t need an explanatory icon at all, IMO.

I saw the header as well. Not sure about that one. IMO, a “Theme Builder” is a different tool, so maybe “Theme Settings”?
I’m not a native English speaker, so label wording is always a hard task if you can't "feel" the language.

ethancrawford:
Final result:
Screenshot 2025-06-11 140921

Issue #2381, opened 2025-05-16
[BUG] Docs search not adjusted to dark mode


BGaweda:
### Browser

N/A

### OS

N/A

### Screen size

N/A

### Describe the bug

Documentation search is light even when the selected theme of Tabler Documentation is dark. 
The text inside the search input is barely visible.

Hovered search input   
![Image](https://github.com/user-attachments/assets/cd45e3c9-cdde-4ab1-bfcf-6fac40d06134)

Search modal:  
![Image](https://github.com/user-attachments/assets/0f58b7a0-bb09-43e1-9dd3-43a5cef3cb78)

### How to reproduce

1. Go to any Tabler Documentation and select a dark theme, e.g.: http://localhost:3010/ui/components/?theme=dark
2. Hover the search input
3. Click the search input and type in any phrase
4. See that the background is white and the font color is very light

ethancrawford:
I notice that docsearch css has variables defined for dark mode. Even so, the colours are a little different to Tabler's. I guess either we'd add `data-theme="dark"` to `<html>` and then override them, or just redefine those variables for `data-bs-theme="dark"`?
https://github.com/algolia/docsearch/blob/main/packages/docsearch-css/src/_variables.css

koraytuncer:
> I notice that docsearch css has variables defined for dark mode. Even so, the colours are a little different to Tabler's. I guess either we'd add `data-theme="dark"` to `<html>` and then override them, or just redefine those variables for `data-bs-theme="dark"`? https://github.com/algolia/docsearch/blob/main/packages/docsearch-css/src/_variables.css

Since Algolia is an extra plugin, it is a bit difficult to intervene in it, so it needs to be customized with .scss.

PR #2443, opened 2025-07-09
fix : Docs search not adjusted to dark mode (#2381)

koraytuncer:
The problem was solved by making various customizations in docs.scss

### DIFF
--- a/.changeset/tricky-moons-laugh.md
+++ b/.changeset/tricky-moons-laugh.md
@@ -1,0 +1,5 @@
+---
+"@tabler/docs": patch
+---
+
+Fix Docs search in dark mode


koraytuncer:
> @koraytuncer I can't speak for @codecalm of course but I suspect that we would want to avoid adding that many custom styles just to get dark mode working. If you have a look at the original issue (#2381) I suggest a few ideas, but again I'm not sure exactly which one of those is the better option if any. Perhaps @BG-Software-BG has some suggestions?

Since Algolia is an extra plugin, it is a bit difficult to intervene in it, so it needs to be customized with .scss.

codecalm (approved):

Issue #2365, opened 2025-05-12
[BUG] Some forms on the new 'Form Elements' page don't lay out neatly


ethancrawford:
### Browser

N/A

### OS

N/A

### Screen size

2560 x 1440

### Describe the bug

After #2303, when viewing the new [form-elements.html](https://github.com/tabler/tabler/blob/dev/preview/pages/form-elements.html) page on a large screen, some of the forms do not line up neatly, due to out of date layout code still expecting a missing column.

### How to reproduce

1. Build and run a local copy of the preview site using any commit from 79bd867db53e49e317aba5de56fe8020e9b48dcc onwards
2. Visit `http://127.0.0.1:3000/form-elements.html`
3. Scroll down to see the 'Basic Form'
4. Notice that all of the following forms don't pack neatly into the layout, but have large gaps

### Screenshots

Current page (not yet deployed to production):
![Image](https://github.com/user-attachments/assets/1365c498-5b3a-489e-8539-94e2669a58cd)

With a potential fix:
![Image](https://github.com/user-attachments/assets/19d28ce1-b221-4b9b-93ad-1b78910caa31)

The potential fix demonstrated in the second gif above is possibly not _quite_ there, as there is still a height difference between columns in the row of three cards. **Should these be forced to the same height?**

I am not sure about how the `row-cards` class works, but the second gif above shows a fix where this class was removed. More testing may be needed to check if it should be left in, but initial tests seemed to work...

ethancrawford:
Fixed with the next deploy when https://github.com/tabler/tabler/pull/2444 goes live.

PR #2444, opened 2025-07-09
fix: Some forms on the new 'Form Elements' page don't lay out neatly …

koraytuncer:
The problem is entirely due to the column structure. It has been solved and has a nice appearance. …(#2365)


koraytuncer:
> @koraytuncer it's helpful if you are able to provide screenshots of before and after the change - that way we don't necessarily have to wait until someone approves the github build workflow to see the result...

I added the images. They are perfectly responsive.
Ekran görüntüsü 2025-07-09 143434

PR #2491, opened 2025-09-17
Blog pages

codecalm:


codecalm:
Closing as no longer applicable — this adds `preview/pages/blog/index.html` and `shared/layouts/blog.html`, both from the pre-Astro Eleventy structure (removed in #2694). GitHub shows this as technically mergeable, but the old layout has no effect in the current Astro-based build. Feel free to reopen against `preview/pages/blog/*.astro` + `shared/astro/` if the blog pages are still wanted.

PR #2497, opened 2025-09-21
Fix deprecated install command for Yarn in docs

Kam0797:
The 'install' command in Yarn has been deprecated in favor of 'add'. Added "install" field to items in package-managers.json and manager-specific commands can be set up here ('add' for Yarn). Updated tabs.package.html accordingly, so the copy field shows 'add' for yarn command


kaiwiessner:
doesn't work, UI is still showing yarn install instead of yarn add ?

ethancrawford:
@kaiwiessner - presumably because while the fix has been merged into dev, the change has not actually been _deployed_...

Kam0797:
I thought the same... Where is it deployed from ? `main`?

ethancrawford:
@Kam0797 I don't know for sure, but I would assume so.

PR #2505, opened 2025-10-06
fix: add border-top-radius to card table first/last cells

### LINKED ISSUE #2493: [BUG] Border issues on some th elements
### Browser

N/A

### OS

N/A

### Screen size

N/A

### Describe the bug

I can see on the 'tables' page of the preview website that there is an issue with th elements on tables that do not have card headers or card body elements above the table. (See screenshot below).
Where this is the case, the border on the outer-most th elements (left and right) seems to fade out, instead of being solid all the way through.
I'm not 100% sure what causes it to occur, but I can also see that if I make the background-color of these th elements transparent, the border is restored properly.

### How to reproduce

1. Go to https://preview.tabler.io/tables.html
2. Observe any tables that have thead and th elements, but do not have card headers or card bodies above them
4. See the fading th border on the outer-most th elements.

### Screenshots

Image

### JSFiddle

_No response_

aastha-cse:
Fixes #2493

Screenshot 2025-10-06 211514


kevinpapst:
@codecalm can you please take care of this? I'd argue that tables are pretty important elements in Admin templates 😁 

The bug is pretty obvious on https://preview.tabler.io/tables.html:

Bildschirmfoto 2025-11-18 um 09 45 08
Bildschirmfoto 2025-11-18 um 09 45 16

kevinpapst (approved):
Tested and works in my environment with 1.4 👍

Issue #2543, opened 2025-11-20
[BUG] Dark-Mode: Checkbox in fieldset not visible


kevinpapst:
### Describe the bug

In a browser it is even less visible than this enlarged screenshot

<img width="1332" height="656" alt="Image" src="https://github.com/user-attachments/assets/02acf96a-feff-41f2-bac3-fdbd40831061" />

See https://preview.tabler.io/form-elements.html?theme=dark

codecalm:
@Bartosz-Do is picking this up. PR incoming soon.

Bartosz-Do:
I fixed this bug in https://github.com/tabler/tabler/pull/2728

kevinpapst:
Thanks @Bartosz-Do 👍

PR #2546, opened 2025-11-23
fix: update dark theme SCSS variables

codecalm:
## Description

This PR fixes the visibility issue of form checkboxes (and other form elements) in dark mode by updating the translucent border color variable. This addresses issue #2543.

## Problem

In dark mode, checkboxes inside `.form-fieldset` elements were barely visible or invisible because the translucent border color (`$border-color-translucent-dark`) had insufficient contrast. The previous value `rgba(72, 110, 149, 0.14)` was too dark and transparent, making form elements hard to see against dark backgrounds.

### DIFF
--- a/.changeset/fix-border-color-translucent-dark.md
+++ b/.changeset/fix-border-color-translucent-dark.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Updated `$border-color-translucent-dark` from `rgba(72, 110, 149, 0.14)` to `rgba(128, 150, 172, 0.2)` to improve visibility of form checkboxes and other form elements in dark mode.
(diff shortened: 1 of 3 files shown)


kevinpapst:
[Before](https://preview.tabler.io/form-elements.html?theme=dark&theme-base=neutral&theme-radius=1):
image

[After](https://tabler-git-fix-2543-tabler-io.vercel.app/form-elements.html?theme=dark&theme-base=neutral):
image

Still a little bit hard to see, but much better than before.

kevinpapst:
Uh oh, I forgot to send this comment and now you merged @codecalm.

I noticed that this changes more borders than just the checkbox, especially the card borders are affected.
And maybe that just me, but in my opinion they looked better before, less prominent.

Maybe this special case of dark theme with even darker fieldset would deserve a special css rule instead of globlaly changing the border? So a conditional rule like
```
theme=dark fieldset { 
    $border-color-translucent-dark: rgba(128, 150, 172, 0.2) !default;
}
```

Left before and right after your change 
Bildschirmfoto 2025-11-25 um 12 52 00
Bildschirmfoto 2025-11-25 um 12 52 10

PR #2553, opened 2025-11-26
Remove redundant nullish coalescing operators in popover and tooltip

Bezlepkin:
The left side always returns a boolean value, the ?? operator is not needed.

This was pointed out by the Angular compiler:


kevinpapst (approved):
Agree, the left side should never be `null`

codecalm:
thank you for help! 🙂

Issue #2324, opened 2025-05-05
[BUG] Pie chart demo's legend cut off on preview page


ethancrawford:
### Browser

N/A

### OS

N/A

### Screen size

N/A

### Describe the bug

Just as the title says.

This is an upstream bug, see link below.

### How to reproduce

1. Go to https://preview.tabler.io/charts.html
2. Scroll down to pie chart example
3. Observe that the legend is cut off.

### Screenshots

![Image](https://github.com/user-attachments/assets/33013aeb-6816-4389-8309-d1c42bb6e093)

ethancrawford:
This appears to be an upstream bug - details at https://github.com/apexcharts/apexcharts.js/issues/4875.

codecalm:
I checked this again on the current `dev` build and cannot reproduce it anymore.

On the charts page the legend of the pie chart and of the polar area chart now fits inside the chart. The last legend row ends 1px above the bottom edge of the 240px canvas, and the legend does not scroll. I measured it at a desktop width and at 375px, where the legend wraps.

The demo now runs on a newer ApexCharts than when this was reported, which most likely explains the difference. I did not check the upstream issue.

Closing as fixed. If you still see the legend cut off, please reopen with the browser and the viewport width.

PR #2555, opened 2025-11-27
feat: upgrade `apexcharts` to v5 and add CSS variables for dynamic chart colors

ethancrawford:
This commit addresses #2324 by upgrading apexcharts and fixing a breaking change by pre-computing some css variables to use in the apexcharts config.

Charts demo page, before upgrading apexcharts: (Note graph at top left):
image

After upgrading apexcharts:
image


codecalm:
Oh, apexcharts can be styled via CSS? So nice! I can help you with upgrade it

Issue #2112, opened 2025-01-30
[BUG] Docker image build process broken?


ethancrawford:
### Browser

N/A

### OS

Windows 11

### Screen size

N/A

### Describe the bug

I've previously been able to run up the tabler docker image and edit/update the source code in real time as directed in the README.
However, after recently messing about with my local copy, I've since been unable to successfully build and run the docker image.
I'll admit that docker is still not something I'm 100% familiar with, so I may be missing something important, but I thought I knew at least enough to manage the basics - so I'm a little bit stumped as to the cause of the error I'm ending up with.
It manages to trigger `gulp start`, but then fails during this task a little bit later with the following:
```
tabler  | [08:37:13] Starting 'build-eleventy'...
tabler  |
tabler  | > @tabler/core@1.0.0 build:html /app
tabler  | > eleventy
tabler  |
tabler  | [11ty] Problem writing Eleventy templates:
tabler  | [11ty] 1. Problem creating an Eleventy Layout for the "./src/pages/2-step-verification-code.html" template file. (via EleventyBaseError)
tabler  | [11ty] 2. You’re trying to use a layout that does not exist: _includes/single (via `layout: single`)
```

See below in 'How to reproduce' for more about what was going on, and questions I have...

### How to reproduce

How I got here:
_Before I started having these issues, I had been dealing with updating my local copy of the `dev` branch, after running into an annoying 'diverging branches' problem, and had assumed I replaced my local `dev` branch with a fresh copy from origin based on the latest from the tabler repo. Although my guess is that this is (probably?) unrelated, I will also admit that perhaps I'm unaware of something in the process that then caused the docker image build to break._ 

_In any case, after apparently updating my local copy of `dev` to 9a46a8fbf9ca0eecca94994dd276c3046f91448f:_

[...]

PR #2599, opened 2026-01-16
Update Dockerfile and docker-compose.yml structure

instance-pro:
The Dockerfile and docker-compose.yml have been updated to match the new project structure.
After starting with `docker compose up --build`, the preview is available at http://localhost:3000 and the docs at http://localhost:3010.

Issue: https://github.com/tabler/tabler/issues/2112

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,15 +1,12 @@
----
-version: "3.3"
 services:
 
   tabler:
     image: tabler
     container_name: tabler
     build: .
+    environment:
+      - NODE_OPTIONS=--no-deprecation
     ports:
       - 3000:3000
-      - 3001:3001
-    volumes:
-      - ./src:/app/src
-      - ./_config.yml:/app/_config.yml
+      - 3010:3010
     tty: true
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,15 +1,12 @@
...
   tabler:
     image: tabler
     container_name: tabler
     build: .
+    environment:
+      - NODE_OPTIONS=--no-deprecation
     ports:
       - 3000:3000
-      - 3001:3001
-    volumes:
```
ihmels, line 11: Why did you remove the volumes? Now I have to rebuild the container image every time I make a change in the source directories (`core/`, `preview/`, `docs/`, `shared/`).

codecalm:
Thanks for the PR! Closing this as no longer applicable — the Dockerfile and docker-compose.yml have since been reworked from scratch in #2089 as part of the pnpm workspaces / Turborepo migration, so this diff (still based on the old `ruby:3.2-alpine` image) no longer applies to the current structure.

For the record, the current setup already addresses the volume-mount concern raised above: docker-compose.yml mounts `core/`, `preview/`, `docs/`, and `shared/` individually and excludes their `node_modules`, so source changes are picked up without rebuilding the image.

Feel free to reopen with a fresh PR against the current Dockerfile if you spot something that still needs fixing there.

PR #2614, opened 2026-02-11
Replace outdated font-weight- classes with valid fw- classes

vitek-dev:
Some demo screens are using outdated font-weight-* classes. Replaced with valid fw-* classes.


codecalm:
thank you for help! 🙂

PR #2618, opened 2026-02-24
Fix `.form-select` box-shadow inside `.input-group`

ionutantohi:
## Summary

- `.input-group` already resets `box-shadow` to `none` on `.form-control` and `.btn`, so those don't show a stray focus/default shadow when grouped. `.form-select` was missing from that list, so a select inside an input-group kept its own box-shadow, looking inconsistent next to the other grouped controls.
- Added `.form-select` to the existing selector.

## Preview

- Vercel needs a maintainer to authorize this external contributor's deployment first (see the bot comment on this PR). Once authorized: find an `.input-group` containing a `.form-select` (e.g. on the form-elements preview page) and confirm it no longer shows a box-shadow, matching the `.form-control`/`.btn` siblings in the same group.

### DIFF
--- a/.changeset/fix-form-select-input-group-shadow.md
+++ b/.changeset/fix-form-select-input-group-shadow.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Fixed `.form-select` keeping its default box-shadow inside `.input-group` by adding it to the existing `.form-control`/`.btn` box-shadow reset.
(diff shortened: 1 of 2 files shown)


ethancrawford:
@ionutantohi - is this in connection to an existing issue?

ionutantohi:
yes. 

a .form-select element still has the box shadow when is used within .input-grup.
image

after the fix
image

PR #2636, opened 2026-04-01
fix: inverted-plus accordion not working with SVG sprite icons

AdilSaifi04:
The inverted-plus accordion toggle relies on targeting internal SVG <path> elements,
which does not work when using SVG sprites (<use>).

This fix replaces path-based manipulation with transform-based rotation,
ensuring compatibility with both inline SVG and sprite icons.

### DIFF
--- a/core/scss/ui/_accordion.scss
+++ b/core/scss/ui/_accordion.scss
@@ -96,9 +96,15 @@
 }
 
 .accordion-button-toggle-plus {
+  .icon,
+  svg {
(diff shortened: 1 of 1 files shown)


codecalm:
Thanks for the fix and for spotting this! The same approach landed on `dev` in #3075, which rotates the toggle element so the plus works with inline SVG, sprite `<use>` and icon fonts. It ships in 1.6. Closing this one as superseded.

PR #2641, opened 2026-04-13
Fix removing data-bs- attributes when already in server-side code.

### LINKED ISSUE #2640: [BUG] tabler flickers in dark mode at page reload dur to removing the HTML attribute
### Browser

Firefox, Chrome,...

### OS

Linux (Kubuntu 24.10/ Neon)

### Screen size

1920x1600

### Describe the bug

when a tabler enabled site is in dark mode (i am using Django), and the HTML attribute `data-bs-theme="dark"` is set server side (comes with the HTML pre-rendered), tabler-theme.js reapplies it after deleting it, when in localstorage it is True.
This leads to a flicker at each page reload. first the page is rendered in light, then switched to dark.

I first thought this is because of my Django site where I use tabler.
But then I could see it on YOUR page, tabler.io, at the preview too.

### How to reproduce

* Enable dark mode on tabler.io
* click on a few links

-> you can see the page flicker, because it first is rendered light, then switch to dark.

The culprit should be in tabler-theme.js

I wrote an override version that does not have the bug for my software.

### Screenshots

_No response_

### JSFiddle

_No response_

nerdoc:
fixes #2640

### DIFF
--- a/core/js/tabler-theme.ts
+++ b/core/js/tabler-theme.ts
@@ -25,19 +25,22 @@ const params = new Proxy(new URLSearchParams(window.location.search), {
 
 for (const key in themeConfig) {
   const param = params[key]
+  const attrName = 'data-bs-' + key
   let selectedValue: string
 
   if (!!param) {
     localStorage.setItem('tabler-' + key, param)
     selectedValue = param
   } else {
-    const storedTheme = localStorage.getItem('tabler-' + key)
-    selectedValue = storedTheme ? storedTheme : themeConfig[key as keyof ThemeConfig]
+ 	const storedTheme = localStorage.getItem('tabler-' + key)
+	// Respect server-rendered attribute when localStorage is empty.
+	const serverValue = document.documentElement.getAttribute(attrName)
+	selectedValue = storedTheme ?? serverValue ?? themeConfig[key as keyof ThemeConfig]
   }
 
   if (selectedValue !== themeConfig[key as keyof ThemeConfig]) {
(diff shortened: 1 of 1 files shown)


Bartosz-Do:
Thanks for tracking down the flicker — the `serverValue` fallback is a good fix for the root cause in #2640. However, the removal branch looks broken:

```js
if (selectedValue !== themeConfig[key as keyof ThemeConfig]) {
  document.documentElement.setAttribute(attrName, selectedValue)
} else if (!document.documentElement.hasAttribute(attrName)) {
  document.documentElement.removeAttribute(attrName)
}
```

This is a no-op either way: if the attribute exists, `!hasAttribute` is `false` and nothing happens; if it doesn't exist, `removeAttribute` has nothing to remove. So once an attribute like `data-bs-theme="dark"` is set (by the server or by this script), it can never be removed again — switching back to the default theme will leave a stale attribute behind. I think this should just stay `else { document.documentElement.removeAttribute(attrName) }` like the original — the `serverValue` fallback above already covers the case this issue was about.

Also a couple of nits:
- Lines 36-39 mix tabs and spaces, inconsistent with the rest of the file.
- Might be worth adding a changeset per the bot's note.

codecalm:
Thanks for this, and sorry it sat here for so long.

The `serverValue` fallback from your PR is what shipped in 1.5.0. It landed on `dev` as part of #2936, in `core/js/tabler-theme.ts`. The removal branch stayed as it was, for the reason Bartosz described above.

Since the fix is already published, I'm closing this instead of rebasing it. #2640 is closed as fixed too.

Issue #2307, opened 2025-04-30
[BUG] Alerts with icons (in docs) spacing issue


HoofedEar:
### Browser

Firefox 138.0 (64-bit)

### OS

Windows 11

### Screen size

1920x1080

### Describe the bug

The spacing on this part of the documentation regarding alerts looks like it is not spacing things properly like the other examples above it? 

### How to reproduce

1. Go to https://docs.tabler.io/ui/components/alerts
2. Scroll down to "Alerts with icons" section

### Screenshots

![Image](https://github.com/user-attachments/assets/05482fee-14de-4813-b4b6-7d055d3018a4)

ethancrawford:
This issue is occurring due to a css property introduced for styling markdown blocks overriding the `line-height` of the alert and its child elements. The line which is interfering is specified here, updated as part of #2219:
https://github.com/tabler/tabler/blob/e6753898714306509e5cdb4e566897b387be3e12/core/scss/ui/_markdown.scss#L6
If this line was not present, the alert would just inherit `line-height` from `body` as normal.

I'm not sure what the desired solution is here - possibly adding a rule like this to [/core/scss/ui/_markdown.scss](https://github.com/tabler/tabler/blob/dev/core/scss/ui/_markdown.scss), or to [/core/scss/ui/_alerts.scss](https://github.com/tabler/tabler/blob/dev/core/scss/ui/_alerts.scss) ?:
```
.markdown .alert {
  line-height: var(--#{$prefix}body-line-height);
}

```

Of course this also raises other questions, such as:
- whether there are other child elements that might appear in a markdown block that should not inherit its `line-height` (though that might be dealt with at a later stage if there's any that arise?) 
- whether the custom line-height for markdown blocks should be applied to **text-based direct child elements** such as `.markdown > p` etc, rather than to the parent `.markdown` element
- whether markdown blocks should be assumed to not contain child elements that need their own custom line-height (so in the case of the alerts documentation page, instead of one big markdown block, the text sections before and after each alert example are contained in separate markdown blocks that do not enclose the live alert demos)

HoofedEar:
I'm curious what the icon section is trying to convey that makes it so different than the first set of alerts? 
Here is the diff between the two sections of code provided. Left side is the first set of alerts, right side is the "icons" section of alerts.

<img width="1322" alt="Image" src="https://github.com/user-attachments/assets/f76443f6-f60e-4415-aad1-988967c23622" />

PR #2649, opened 2026-05-01
[BUG] Alerts with icons (in docs) spacing issue #2307

resul-elezi:
**This pull request makes the following changes:**

*Fixes issue #2307 

The alert examples in the "Alerts with icons" section were missing the intended visual spacing (gap) between the icon and the content.

Changes:

- Restructured Alert HTML: Removed the redundant d-flex wrapper inside the alert components.
- Renamed Classes: Renamed the wrapper from d-flex to alert-icon to align with Tabler's standard markup.
- Flattened DOM Hierarchy: By removing the extra nesting level, the icon and text content are now direct children of the alert container.

Why this approach?

Tabler alerts use Flexbox on the parent container to manage spacing via the gap property. The previous nesting prevented this gap from affecting the icons. Instead of introducing utility classes to fix the symptoms, this change restores the originally intended structure. This ensures the design is consistent with the "Default markup" and maintains long-term maintainability.

### DIFF
--- a/docs/content/ui/components/alert.md
+++ b/docs/content/ui/components/alert.md
@@ -51,33 +51,32 @@ Use the `alert-icon` class on an `<svg>` (or on an `<i>` when using the webfont)
 
 {% capture html -%}
 <div class="alert alert-success" role="alert">
-  <div class="d-flex">
-    <div>
+   <div class="alert-icon">
+      <!-- Download SVG icon from http://tabler.io/icons/icon/check -->
       <svg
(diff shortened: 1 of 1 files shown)


codecalm:
Closing as no longer applicable — this touches `docs/content/ui/components/alert.md`, a path/format removed by the Eleventy → Astro migration (#2694); the content now lives at `docs/pages/ui/components/alert.mdx`. Feel free to reopen against the current file if the alert-icon spacing issue still reproduces.

PR #2672, opened 2026-06-30
Improve form control border contrast to meet WCAG 1.4.11

f2cmb:
While checking form accessibility and making a compliance work on https://github.com/glpi-project/glpi , 

I noticed the default field borders are really hard to see on a light background. The input and select border uses `--tblr-border-color` (`gray-200`, `#e5e7eb`) and the checkbox/radio border uses `--tblr-border-color-translucent`, and both sit at roughly 1.2:1 against white. That's well under the 3:1 that WCAG 2.1 SC 1.4.11 asks for, so an empty field or an unchecked control can be tricky to make out.

The fix derives those borders from the text color rather than from gray-200, through a new token:

```scss
$form-control-border-color: color-mix(in srgb, var(--#{$prefix}body-color), transparent 40%) !default;
```

Because `body-color` flips with the theme, this lands around 3.3:1 on the default light palette and around 6:1 on dark, without anything hardcoded per theme. `$input-border-color` (so selects and input-group addons follow along) and the `form-check` border tokens now point at it.

A few notes on scope:

* It targets the controls whose border is actually drawn in the default theme: inputs, selects, checkboxes and radios.
* The checked checkbox/radio is identified by its primary fill and has no border in the default light theme, so there's no contrast regression there. I still repointed the checked-border token so themes that do render that border stay consistent.
* I left `$border-color` alone on purpose, so cards, tables and dividers are untouched.

Quick before/after: on `main` an empty input border measures about 1.2:1 against the surface; with this change the same border comes out around 3.3:1, so the box is clearly visible.

Live repro: https://jsfiddle.net/xvq5zbkj/

A changeset is included.

### DIFF
--- a/.changeset/form-control-border-contrast.md
+++ b/.changeset/form-control-border-contrast.md
@@ -1,0 +1,7 @@
+---
+"@tabler/core": patch
+---
+
+Derived form control and checkbox/radio borders from the text color token so the
+control boundary reaches the 3:1 non-text contrast required by WCAG 1.4.11. The
+previous `gray-200` border sat at ~1.2:1 on light surfaces.
--- a/core/scss/_variables.scss
+++ b/core/scss/_variables.scss
@@ -1791,7 +1791,10 @@ $list-group-action-active-bg: var(--#{$prefix}secondary-bg) !default;
 
 // Forms
 $input-disabled-bg: $disabled-bg !default;
-$input-border-color: var(--#{$prefix}border-color) !default;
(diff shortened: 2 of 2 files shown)


ethancrawford:
Perhaps a bigger discussion on a new issue, but besides improving the visibility of some of the default elements, I think it would be helpful to have the option of theming Tabler UI with a bunch of accessible colour themes, (colour-vision-deficiency-friendly options, plus high contrast maybe). @Bartosz-Do / @codecalm what say you? What scope is there for adding this?

PR #2681, opened 2026-07-07
Fix `.bg-gradient` conflict and add gradient documentation pages

Bartosz-Do:
## Summary

- Remove the duplicate utility entry that overrode `.bg-gradient` behavior in core SCSS.
- Restore utility-based gradients that use `from`/`via`/`to` color stops.
- Add new docs pages for `Card gradient` and `Gradients` with examples and usage patterns.
- Add a changeset for `@tabler/core` and `@tabler/docs` patch releases.

## Preview

- **How to test:** Open both pages, check that gradient examples render correctly, then verify direction classes (`.bg-gradient-to-*`) and color stop classes (`.bg-gradient-from-*`, `.bg-gradient-via-*`, `.bg-gradient-to-*`) visibly change the output.

### DIFF
--- a/core/scss/_utilities.scss
+++ b/core/scss/_utilities.scss
@@ -794,13 +794,6 @@ $utilities: map.merge(
       class: bg,
       values: $utilities-bg-subtle,
     ),
-    'gradient': (
-      property: background-image,
-      class: bg,
-      values: (
-        gradient: var(--#{$prefix}gradient),
-      ),
-    ),
     'user-select': (
       property: user-select,
       values: all auto none,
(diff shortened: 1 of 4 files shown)


core/scss/_utilities.scss
```diff
@@ -794,13 +794,6 @@ $utilities: map.merge(
       class: bg,
       values: $utilities-bg-subtle,
     ),
-    'gradient': (
```
codecalm, line 797: why do you remove it from css?

Bartosz-Do: I removed it because it creates a duplicate .bg-gradient class.
That duplicate can override the new from/via/to gradient rules and break consistency.

When it wasn't deleted gradients on docs were not working. I asked AI why and it found it.

PR #2740, opened 2026-07-30
Fix invalid Sass if() syntax breaking the core build

codecalm:
## Summary

- Restored valid Sass `if(cond, a, b)` syntax in place of the invalid `if(sass(cond): a; else: b)` form across 22 files in `core/scss`, which broke Sass compilation entirely (`sass` exited with a parse error on `core/scss/mixins/_functions.scss:44` and would fail identically on the other 21 files once reached).
- No behavior change: each replacement is a pure syntax fix, same condition/true/false branches, verified by diffing before/after compiled CSS output.

## Preview

- **How to test:** No visual change (build-fix only). Run `pnpm --dir core run css` (or `sass --load-path=node_modules core/scss/:...`) from a clean checkout of `dev` to see it fail, then on this branch to see it succeed with only Sass's built-in `if()`-deprecation warnings (no errors).

## Notes / rollout

- Root cause: an apparent mass find/replace in `9d5c83f3a` ("Remove Bootstrap dependency and vendor JS/SCSS sources into Tabler") introduced the invalid syntax across ~70 call sites in 22 files.

### DIFF
--- a/.changeset/fix-sass-if-function-syntax.md
+++ b/.changeset/fix-sass-if-function-syntax.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Fixed invalid Sass `if(sass(cond): a; else: b)` syntax back to the correct `if(cond, a, b)` form across `core/scss`, which was breaking the Sass build.
--- a/core/scss/_maps.scss
+++ b/core/scss/_maps.scss
@@ -154,6 +154,6 @@ $utilities-border-subtle: (
 
 $utilities-links-underline: map-loop($utilities-colors, rgba-css-var, '$key', 'link-underline') !default;
 
-$negative-spacers: if(sass($enable-negative-margins): negativify-map($spacers); else: null) !default;
+$negative-spacers: if($enable-negative-margins, negativify-map($spacers), null) !default;
 
 $gutters: $spacers !default;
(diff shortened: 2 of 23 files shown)


codecalm:
I compiled both branches locally with the project's own Dart Sass (1.102.0, `core/node_modules/.bin/sass`) to check this.

**`dev` (current):** uses `if(sass(cond): a; else: b)` — the modern CSS-native `if()` syntax that Dart Sass added as the non-deprecated replacement for the classic global `if()` function. Compiles with **0 errors, 0 warnings**.

**This PR:** reverts those call sites to the classic `if(cond, a, b)` form. That form still works, but it's exactly what's now deprecated — compiling this branch produces ~80 `DEPRECATION WARNING [if-function]: The Sass if() syntax is deprecated in favor of the modern CSS syntax` warnings, one per call site touched here (see https://sass-lang.com/d/if-function).

So the premise ("invalid Sass `if()` syntax... broke Sass compilation entirely") doesn't hold against current `dev` — `dev` already compiles cleanly using the syntax this PR removes. Merging this would trade a clean, warning-free build for one with dozens of deprecation warnings.

Given that, I don't think this should be merged as-is. If there's a real compilation failure this was meant to fix, it'd help to see the actual error output/environment, since it doesn't reproduce here.

PR #2749, opened 2026-08-01
Add quality gates: Prettier for Astro, astro check for shared, unused SCSS variable lint

codecalm:
## Summary

- Format `.astro` files with `prettier-plugin-astro` and widen the lint/format globs to cover Astro components and build scripts. 12 files that mix HTML comments or `<script>` inside JSX are excluded (the plugin can't parse those yet) and listed in `.prettierignore`.
- Switch `@tabler/shared` type-check from `tsc --noEmit` to `astro check`. This found 46 real type errors across 15 components; all are fixed without changing the rendered HTML.
- Move `DocsLayout` and `DocsMdxLayout` from `shared/layouts` to `docs/layouts`, since they only ever import docs-only components. Add a `@layouts` alias and update the `layout:` front matter in all docs pages.
- Wire `find-unused-sass-variables` into the root `lint` script and remove the unused `!default` variables it found. The color shade ladders (`$blue-100`…`$cyan-900`) and per-color maps stay, marked with `fusv-disable` since they are meant to be configurable even if unused internally.
- Fix a dev-mode race in `docs`/`preview`'s `copy-assets` scripts: wait for `@tabler/core`'s first build instead of failing on a fresh clone.

## Preview

- **How to test:** This change only touches formatting, types, and internal lint config — no visible UI changed. Verified with a canonical DOM diff of every built `preview` and `docs` page against the pre-change build (all pages match), plus a byte-for-byte diff of `core`'s compiled CSS.
- To sanity-check: `pnpm run lint`, `pnpm --filter @tabler/shared type-check`, `pnpm --filter @tabler/preview build`, `pnpm --filter @tabler/docs build`.

## Notes / rollout

- The SCSS variable removal is a breaking change for anyone using `@use '@tabler/core' with (...)` to override one of the removed variables (all were dead — Bootstrap-deprecated aliases, unused icon/scale variables — with no effect on compiled CSS). Included as a `major` changeset for `@tabler/core`.


codecalm:
@copilot resolve the merge conflicts in this pull request

PR #2803, opened 2026-08-06
fix: resolve Windows cmd.exe build errors

ankitsinghbash:
### What does this PR do?
This PR fixes two critical issues that break the build process on Windows environments using `cmd.exe`:

1. **Terser Pipe Syntax Error (`@tabler/core`, `@tabler/preview`)**: 
   The `--comments` flag in `package.json` used `'/@license|@preserve|^!/'`. In Windows `cmd.exe`, the pipe character (`|`) is parsed as a shell pipe, causing the build to fail with `'preserve' is not recognized as an internal or external command`. Since `terser`'s default behavior already preserves `@license`, `@preserve`, and `!` comments, passing the regex explicitly was unnecessary. I simplified this by removing the explicit regex, which safely relies on Terser's defaults.
2. **`npx` EINVAL Error on Node 20.12+ (`@tabler/preview`)**: 
   The `prettify-html` Astro hook previously spawned `npx` via `execFileSync`. With the recent `BatBadBut` security mitigation introduced in Node v20.12+, spawning `.cmd` binaries without `shell: true` immediately throws an `EINVAL` error on Windows. I refactored the execution to use `execSync` with a single command string, which gracefully bypasses the deprecation warnings and safely leverages `cmd.exe` execution across all platforms.

### Test Plan
Ran `npm run build` locally on Windows. Both `@tabler/core` and `@tabler/preview` now compile successfully without throwing `ENOENT` or `EINVAL` errors.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -5,6 +5,7 @@
   "homepage": "https://tabler.io",
   "scripts": {
     "dev": "pnpm run watch",
+    "test-escape": "concurrently \"echo '/@license^|@preserve^|^^!/'\"",
     "dev-prepare": "pnpm run clean && pnpm run copy && concurrently \"pnpm run css-build\" \"pnpm run js-build\"",
     "build": "pnpm run clean && pnpm run build-assets && pnpm run copy && pnpm run generate-sri",
     "build-assets": "concurrently \"pnpm run css\" \"pnpm run js\"",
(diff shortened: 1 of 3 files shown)


Bartosz-Do:
Thanks for digging into this — both root causes are real and correctly diagnosed. I tested the branch locally (macOS, Node 24.17, pnpm 11.17); here's what I found.

**What checks out**

- The `--comments` claim holds. I built `core` on this branch and on `dev` and compared the output: all four minified files are byte-identical, license banner included. I also checked that a bare `--comments` doesn't swallow the following `--source-map` argument — terser's parser handles it, source maps come out fine.
- `preview` builds, the `prettify-html` hook runs, and the exclusions still bite. Across all 126 generated HTML files the only difference vs `dev` is the build timestamp.
- Scope is complete — no other `npx` call sites, no other single-quoted args in `package.json` scripts.

I don't have a Windows machine, so I couldn't verify the cmd.exe build itself.

**Needs fixing before merge**

1. **Leftover debug script.** `core/package.json:8` has `"test-escape": "concurrently \"echo '/@license^|@preserve^|^^!/'\""` — looks like a leftover from experimenting with `^|` escaping. Please drop it.

2. **CI lint will fail.** `preview/astro.config.mjs` isn't formatted to the repo's prettier config, and `lint-prettier` covers `{preview,docs}/**/*.mjs`. `pnpm run format-prettier` fixes it.

3. **Please restore the deleted comment** above `--ignore-path`. It explains why the flag exists at all: `.gitignore` and `.prettierignore` both list `dist`, so without pointing at `devNull` the pass silently no-ops on the very directory it's meant to format. Without that note someone will remove the flag later as "unnecessary" and CI will stay green.

4. **`execSync` with an interpolated string.** I'd rather not reintroduce shell quoting here — `outDir` goes straight into the command, so a `$`, backtick or quote in the project path breaks it. And it isn't necessary: the problem is `npx` resolving to `npx.cmd`, not `execFileSync`. Skipping `npx` and running prettier's CLI through node sidesteps the whole class of issue:

```js
import { createRequire } from 'node:module'
const require = createRequire(import.meta.url)

execFileSync(
  process.execPath,
  [require.resolve('prettier/bin/prettier.cjs'), '--write', '--parser', 'html', /* ...unchanged... */],
  { stdio: 'inherit' },
)
```

No shell, no `.cmd`, works everywhere, slightly faster, and it keeps both the args array and the comment from (3). I'd avoid plain `{ shell: true }` on `execFileSync` — on POSIX, Node joins the args unquoted, so the shell would expand the globs before prettier sees them.

**Minor**

- With the regex gone, the bare `--comments` is a no-op that just restates the default — cleaner to drop the flag entirely (verified: identical output either way).
- Pre-existing rather than yours, but relevant to the goal: on Windows `outDir` has `\` separators, which globs treat as escapes. Worth confirming the prettier pass actually formats something on Windows, not just that it stopped crashing.
- Branch is two commits behind `dev` — worth a rebase.

Thanks again for the fix!

codecalm:
Thank you for the report and the fix. Both issues are gone on `dev` now: #2936 removed the `--comments` flag from every Terser call, and the `prettify-html` integration runs prettier in worker threads instead of spawning `npx`, so the Windows build no longer hits either error. Closing since there is nothing left to merge, the fix ships with 1.5.

PR #2805, opened 2026-08-06
Fix all `pnpm type-check` warnings and restore lost `Props` typing

Bartosz-Do:
## Summary

- `pnpm type-check` now reports 0 errors, 0 warnings and 0 hints in all three packages (`shared`, `preview`, `docs`). It reported 39 warnings before.
- Nine of those warnings hid a real bug. Astro's frontmatter scanner gives up when a `Props` member is named `as`, or when a JSDoc comment inside `Props` contains a `>`. When that happens the compiler falls back to `Record<string, any>`, so the props of those components were never type-checked at call sites. The `as` prop now lives in a separate `PolymorphicProps` type (8 components) and is typed `HTMLTag` instead of `string`. Turning real typing back on surfaced 3 genuine errors in `CardHeader`, `CardTitle` and `FormHint`, which are fixed here.
- Routine cleanup: explicit `is:inline` on 18 `define:vars` scripts (they were already inline), removed unused imports and variables, and added ambient declarations for the vendor globals `Typed`, `ClipboardJS` and `window.TomSelect`.
- Two small behaviour fixes: the advanced table now passes `this` instead of the deprecated global `event`, and the invoice print button drops its `javascript:` prefix (which parsed as an unused label).

## Preview

- **URL:** [preview link](https://tabler-git-fix-type-check-warnings-tabler-io.vercel.app/tables.html)
- **How to test:**
  - `/tables.html` — open the record-count dropdown in the advanced table footer and pick a value. The counter next to the button and the number of rows should both update. This is the one real behaviour change (`setPageListItems(event)` → `setPageListItems(this)`).
  - `/invoice.html` — click "Print Invoice". The print dialog should still open.
  - `/charts.html`, `/maps.html`, `/form-elements.html`, `/marketing/hero.html` — these pages run the scripts that gained `is:inline`. Check the widgets still render and the console is clean.
  - Card, list-group and form-hint components are used across most pages, so a general click-through is worth it after the `Props` change.

## Notes / rollout

- No breaking changes. The public `as` prop keeps its name; only where the type is declared changed.
- The `as` prop could be renamed (for example to `tag`, matching `element` in `Button.astro`) to avoid the workaround entirely. That was left out on purpose: it would change the API of 8 shared components and every call site.
- The workaround is fragile in one specific way, and there is a comment in each file saying so: quoting the key as `'as'` also works, but Prettier strips the quotes (`quoteProps: "consistent"`) and would silently undo the fix.
- Verified with `pnpm type-check`, `pnpm test` (112 passing), `pnpm lint-prettier`, and a browser check of the table dropdown.


codecalm:
@copilot resolve the merge conflicts in this pull request

PR #2847, opened 2026-08-11
Fix `.dropstart` caret rendering after logical-properties migration

### LINKED ISSUE #2821: Incomplete physical-to-logical property migration breaks RTL layouts
`core/scss/mixins/bootstrap/_caret.scss:36,50,67` (called from `core/scss/bootstrap/_dropdown.scss:164` and `_button-group.scss:92-94`) — the `caret()` mixin special-cases `$direction == 'left'` to move the glyph to `::before`, but every real caller now passes the logical direction name `'start'` after the physical→logical rename, so the special case never fires. The caret ends up rendering after the button label (on `::after`) instead of before it, and the dead `::before` rule never produces content. Confirmed independently by two separate review passes and against the compiled CSS.

Bartosz-Do:
## Summary

- The `caret()` Sass mixin still checked for the old `'left'` keyword to move the arrow onto `::before`. After the physical-to-logical property migration, every real caller now passes `'start'` instead, so that check never matched.
- Result: `.dropstart` dropdown toggles (menu opens toward the start edge) rendered their arrow on `::after` — after the label — instead of before it, in both LTR and RTL.
- Fix: change the three `'left'` comparisons in `core/scss/mixins/bootstrap/_caret.scss` to `'start'`. No other files touched.

## Preview

- **URL:** [preview link](https://tabler-git-fix-2821-tabler-io.vercel.app/dropdowns.html)
- **How to test:** The public dropdowns demo page does not include a `.dropstart` example, so the fix won't be visible there. To check it directly: build the core CSS (`pnpm --filter @tabler/core run css-build`) and open `core/js/tests/visual/dropdown.html` in a browser. On the "Dropstart" and "Dropstart split" buttons, the small arrow should sit before the label (left of it in LTR, right of it with an RTL build), not after.

## Notes / rollout

- Fixes #2821.
- CSS-only change, no JS or markup changes.

### DIFF
--- a/core/scss/mixins/bootstrap/_caret.scss
+++ b/core/scss/mixins/bootstrap/_caret.scss
@@ -33,7 +33,7 @@
 @mixin caret($direction: down, $width: $caret-width, $spacing: $caret-spacing, $vertical-align: $caret-vertical-align) {
   $selector: 'after';
 
-  @if $direction == 'left' {
+  @if $direction == 'start' {
     $selector: 'before';
   }
 
@@ -47,7 +47,7 @@
     border-inline-start: 1px var(--border-style);
     margin-inline-end: 0.1em;
 
-    @if $direction != 'left' {
+    @if $direction != 'start' {
       margin-inline-start: $caret-spacing;
     } @else {
       margin-inline-end: $caret-spacing;
@@ -64,7 +64,7 @@
     }
   }
 
-  @if $direction == 'left' {
+  @if $direction == 'start' {
     &::after {
       content: none;
     }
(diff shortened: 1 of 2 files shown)


codecalm:
Superseded by #2848, merged to `dev` as 2c55ff32.

That PR contains the identical `_caret.scss` and `_caret.test.scss` changes (verified byte-for-byte against this branch) plus two additional fixes: the missing `$breakpoints` forwarding in `media-breakpoint-down-than()` and the `--link-opacity` unit mismatch in `color-mix()`.

Closing as a duplicate — the caret fix is shipped. Thanks!

PR #2878, opened 2026-08-14
Add SRI `integrity` hashes to CDN snippets in docs

### LINKED ISSUE #2876: Add SRI integrity hashes to the CDN snippets in the docs
`core/.build/generate-sri.ts` already runs as the last step of `core build` and writes 18 sha384 hashes to `shared/data/sri.json` — for `tabler.min.css`, its RTL twin, every plugin stylesheet, `tabler.min.js` and `tabler-theme.min.js`. Nothing reads that file. The only CDN snippets in the docs are `docs/components/CdnImportPackage.astro` and `docs/components/CdnImportPlugin.astro`, and neither emits `integrity`.

## Steps

- [ ] read `shared/data/sri.json` in `CdnImportPackage.astro` and add `integrity` + `crossorigin="anonymous"` to the CSS and JS tags
- [ ] same for `CdnImportPlugin.astro` — it builds `tabler-{plugin}.min.css` URLs and the matching keys (`css-flags`, `css-payments`, `css-socials`, …) already exist
- [ ] make the docs build safe when the file is missing: `sri.json` is gitignored (`.gitignore:41`) and only exists after `core build`. Check the dependency in `turbo.json` and fail loudly rather than shipping `integrity=""`
- [ ] mention SRI in the installation docs
- [ ] changeset (`@tabler/docs: minor`)

## Release ordering

`site.cdnUrl` builds the URL from the package version, so the snippet and the hash can drift apart: while `core/package.json` still says 1.4.0, the page would show a 1.4.0 URL next to a hash taken from a local 1.5 build. Fix the order — `changeset version` → build → deploy docs — and consider a post-publish check that the hashes match what jsDelivr actually serves.

## Done when

The docs show CDN snippets with valid `integrity`, a browser accepts the resource from jsDelivr without an SRI error, and the docs build fails clearly if `sri.json` is absent instead of emitting an empty attribute.

Bartosz-Do:
## Summary

- CDN snippets in the docs now carry `integrity` and `crossorigin="anonymous"`, so a browser stops a tampered or corrupted file instead of running it.
- Hashes describe the published release the snippets link to. `generate-sri` no longer runs on every `core build` — it runs once per release, in the release workflow, right after `changesets` publishes that exact `core/dist`. Hashing a mid-cycle build would pin a value the browser rejects, because `dev` moves on while the snippets still link to the last released version.
- `shared/data/sri.json` is now committed and carries the release it belongs to (`{ "version": "1.4.0", "hashes": { … } }`). The docs render the hashes only when that version matches `@tabler/core`, so a stale file drops `integrity` instead of shipping a wrong one. A docs build no longer needs a prior core build.
- Every CDN snippet is built in one place (`docs/lib/cdn-snippets.ts`): the doc pages, the `/llms.txt` endpoints, and the full HTML example on the installation page. Before, the last two were separate copies with no `integrity`.

## Preview

- **URL:** [Installation](https://tabler-docs-git-fix-2876-tabler-io.vercel.app/ui/getting-started/installation), [Flags plugin](https://tabler-docs-git-fix-2876-tabler-io.vercel.app/ui/plugins/flags), [installation.md](https://tabler-docs-git-fix-2876-tabler-io.vercel.app/ui/getting-started/installation.md)
- **How to test:** Open the installation page. Both code blocks — the short CDN snippet and the full HTML example — should show `integrity="sha384-…"` and `crossorigin="anonymous"`. The `.md` version of the page (what `/llms.txt` serves) should show the same. Copy the snippet into an empty HTML file and open it: the styles must apply, which is what proves the hash matches the file users get from the CDN.

## Notes / rollout

- `release.yml` gains two steps, both behind `if: steps.changesets.outputs.published == 'true'`: run `generate-sri`, then commit `shared/data/sri.json` back to `dev` as `chore: update SRI hashes [skip ci]`. The checkout now uses `fetch-depth: 0` so that push works. If `dev` is protected against direct pushes, this step needs a token that may bypass it.
- Until that commit lands, the version in `sri.json` does not match `package.json` and the snippets render without `integrity` — safe, just unprotected.
- A hash missing while the version does match fails the docs build. That case means the file list in `core/.build/generate-sri.ts` drifted from what the docs render.
- Closes #2876

### DIFF
--- a/.changeset/cdn-sri-hashes.md
+++ b/.changeset/cdn-sri-hashes.md
@@ -1,0 +1,5 @@
+---
+"@tabler/docs": minor
+---
+
+Added `integrity` and `crossorigin` to the CDN snippets in `CdnImportPackage` and `CdnImportPlugin`.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v7
+        with:
+          fetch-depth: 0 # full history, so the SRI commit below can be pushed back to dev
 
       - name: Install PNPM
         uses: pnpm/action-setup@v6
@@ -39,6 +41,7 @@ jobs:
         run: pnpm run build
 
       - name: Create release Pull Request or publish to NPM
+        id: changesets
         uses: changesets/action@v1
         with:
           version: pnpm run version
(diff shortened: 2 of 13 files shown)


codecalm:
Closing this one — SRI is deferred to v2.0. The generator script is kept, unwired, in #2886, which also lands the CDN snippet helper this PR introduced.

PR #2916, opened 2026-08-31
Add Info, Tip, Warning, Danger and Note callouts to the docs

codecalm:
## Summary

- Adds `Info`, `Tip`, `Warning`, `Danger` and `Note` callouts for docs pages: one `Callout` component with named wrappers, passed to `<Content />` so any MDX page can use the tags without an import.
- Adds a reference page under Resources with the five variants, the usage rules and the accessibility notes.
- Converts four existing asides that were easy to miss in the flow of a page: the Mapbox secret token warning, the Sass `@import` deprecation, the page-layouts prerequisite and the WYSIWYG theme tip.

## Preview

- **URL:** [docs callouts](https://tabler-docs-git-docs-callouts-tabler-io.vercel.app/ui/getting-started/callouts)
- **Also changed:** [map](https://tabler-docs-git-docs-callouts-tabler-io.vercel.app/ui/components/map), [upgrade](https://tabler-docs-git-docs-callouts-tabler-io.vercel.app/ui/getting-started/upgrade), [page layouts](https://tabler-docs-git-docs-callouts-tabler-io.vercel.app/ui/layout/page-layouts), [wysiwyg](https://tabler-docs-git-docs-callouts-tabler-io.vercel.app/ui/plugins/wysiwyg)
- **How to test:** open the callouts page in both color modes and check the five variants. The tinted backgrounds come from the `-lt` color tokens, so they adapt on their own. Each callout renders as `<aside role="note">` with a label built from the variant and the title.

## Notes / rollout

- The styles live in `docs/scss/docs.scss`, not in core: callouts are for documentation pages, not a Tabler UI component. An alert inside a product interface still uses the alert component.
- The color of a variant sits on the icon and the border only. Text keeps the body color, so a pale state like yellow never has to carry text and nothing depends on color alone.

### DIFF
--- a/.changeset/docs-callouts.md
+++ b/.changeset/docs-callouts.md
@@ -1,0 +1,5 @@
+---
+"@tabler/docs": minor
+---
+
+Added `Info`, `Tip`, `Warning`, `Danger` and `Note` callouts, usable in any MDX page, with a reference page under Resources.
--- a/docs/components/Callout.astro
+++ b/docs/components/Callout.astro
@@ -1,0 +1,40 @@
+---
+// A tinted aside for a note, a tip or a warning inside docs prose. MDX pages use the
+// named wrappers next to this file (<Info>, <Tip>, <Warning>, <Danger>, <Note>), which
+// [...slug].astro passes to <Content /> — so a page can call them without importing.
+import Icon from '@ui/Icon.astro'
+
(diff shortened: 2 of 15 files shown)


codecalm:
Follow-up: #2917 uses these callouts for the ApexCharts license note on the chart page.

PR #2991, opened 2026-09-07
Fix `.dropdown-menu-arrow` border color, replace `clip` with `clip-path`

### LINKED ISSUE #2962: [BUG] Dropdown arrows in Firefox
### Tabler version

1.5.0

### Browser

Firefox 155.0.1

### OS

MacOS

### Screen size

1920x1080

### Describe the bug

Wrong color + 2 lines underneath

Image

### How to reproduce

1. Go to https://preview.tabler.io/dropdowns.html
2. Click on top right username or look at dropdown examples with arrows on top

### Screenshots

_No response_

### Live example

_No response_

codecalm:
## Summary

- The `.dropdown-menu-arrow` outline was drawn in the text color. Since #2839 the menu has no `border` (its edge is the 1px ring in `--tblr-shadow-dropdown`), so `border-color: inherit` on the arrow resolved to `currentColor`. The arrow now uses `var(--tblr-dropdown-border-color)`, which also follows `.dropdown-menu-dark` and the dark theme.
- The arrow is cut with `clip-path: inset(0 5px 5px 0)` instead of the deprecated `clip: rect()`. Same shape, same RTL output.

Checked in Chromium and Firefox 150, light and dark. The extra lines from the report could not be reproduced in Firefox 150; the `clip` replacement removes the one deprecated property the arrow relied on.

## Preview

- **URL:** [dropdowns.html](https://tabler-git-fix-dropdown-arrow-border-tabler-io.vercel.app/dropdowns.html)
- **How to test:** open the user menu in the top-right navbar, or look at the two dark dropdowns with an arrow in the right column. The arrow outline matches the menu edge instead of being a dark line. Switch to dark mode and check the same menus.

## Notes / rollout

- Closes #2962

### DIFF
--- a/.changeset/fix-dropdown-arrow-border.md
+++ b/.changeset/fix-dropdown-arrow-border.md
@@ -1,0 +1,5 @@
+---
+"@tabler/core": patch
+---
+
+Fixed the `.dropdown-menu-arrow` border color and replaced the deprecated `clip` with `clip-path`.
--- a/core/scss/ui/_dropdowns.scss
+++ b/core/scss/ui/_dropdowns.scss
@@ -86,11 +86,12 @@
     display: block;
     width: 14px;
     height: 14px;
-    clip: rect(0, 9px, 9px, 0);
+    clip-path: inset(0 5px 5px 0);
     content: '';
     background: inherit;
-    border: 1px solid;
-    border-color: inherit;
+    // The menu draws its edge with the 1px ring in --dropdown-box-shadow and has
+    // no border, so `inherit` would give the arrow the text color.
+    border: 1px solid var(--dropdown-border-color);
 
     /* rtl:ignore */
     transform: rotate(45deg);


ethancrawford:
@codecalm It looks like this is only half fixed - here is a screenshot of part of the preview site running in dev mode on my computer on Firefox 155.0.1. - Note the border colour is now correct, but the extra lines from incorrect clipping are still under the dropdown arrows.

This appears fine in Chrome, so it's possibly a firefox-specific rendering bug?

image

Claude suggested to me that a potential fix might be to replace clip-path altogether by only rendering the parts of the border we want, by doing something like the following:

`core/scss/_variables.scss`:
```diff
-$dropdown-arrow-offset: 0.25rem !default;
-$dropdown-arrow-inset: 0.75rem !default;
+$dropdown-arrow-offset: 0.315rem !default;
+$dropdown-arrow-inset: 0.906rem !default;
```

`core/scss/ui/_dropdowns.scss`:
```diff
.dropdown-menu-arrow {
  &::before {
    position: absolute;
    inset-inline-start: var(--dropdown-arrow-inset);
    top: calc(-1 * var(--dropdown-arrow-offset));
    z-index: -1;
    display: block;
-   width: 14px;
-   height: 14px;
-   clip-path: inset(0 5px 5px 0);
+   width: 9px;
+   height: 9px;
    content: '';
    background: inherit;
    // The menu draws its edge with the 1px ring in --dropdown-box-shadow and has
    // no border, so `inherit` would give the arrow the text color.
-   border: 1px solid var(--dropdown-border-color);
-
+   border-width: 1px 0 0 1px;   // top + left only
+   border-style: solid;
+   border-color: var(--dropdown-border-color);
    /* rtl:ignore */
    transform: rotate(45deg);
    transform-origin: center;
  }
}
```

Haven't properly tested that though, so 🤷‍♂️

PR #3007, opened 2026-09-08
Ensure folders needed by build-css.ts exist when needed

### LINKED ISSUE #3006: [BUG] Build error due to preview css folder not existing when it is referenced
### Tabler version

v1.5

### Browser

N/A

### OS

Windows

### Screen size

N/A

### Describe the bug

See attached for the full build log.

I can reliably reproduce the same `ENOENT` error. I think the `build-css.ts` script currently assumes that `public/preview/css/` exists by the time it gets to write the css files there, but after the `copy-assets.ts` script deletes the public folder, it is perhaps not created again early or quickly enough for `build-css.ts` to use.

[tabler build log.txt](https://github.com/user-attachments/files/31952082/tabler.build.log.txt)

### How to reproduce

On my Windows computer, I can reproduce this by doing the following with the `@tabler/core@1.5.0` tag checked out:
1. Run `pnpm dev` in the root of the project, 
2. Observe the `ENOENT` error in the console's build log output.

### Screenshots

_No response_

### Live example

_No response_

ethancrawford:
Previously, there was a potential race condition between copy-assets.ts and build-css.ts, as one wiped the /public folder and the other just assumed that it was available when writing assets to it. By always making the build-css script run mkdirSync for the required path, we can ensure that the path exists.

There's a fairly detailed comment added to my fix, but I can edit or remove it if preferred. Let me know.

Fixes #3006

### DIFF
--- a/.agents/skills/build-pipeline/SKILL.md
+++ b/.agents/skills/build-pipeline/SKILL.md
@@ -19,7 +19,7 @@ Astro only renders pages. Every stylesheet, script, font and image on a page arr
 | Directory | Written by | Notes |
 | --- | --- | --- |
 | `core/dist/` | `@tabler/core` (`css:dev`, `js:build`, `copy`) | the framework: css, js, fonts, img, vendored libs |
-| `preview/tmp-assets/` | `preview` `assets` (`build-css.ts` + vite/terser) | demo css/js, isolated from Astro's output |
+| `preview/tmp-assets/` | `preview` `assets` / `watch:css` (`build-css.ts` + vite/terser) | demo css/js, isolated from Astro's output |
 | `docs/tmp-assets/css/` | `docs` `css` / `watch:css` (build-css, `--no-prefix`) | docs stylesheets |
 | `<pkg>/public/` | **`copy-assets` only** | fully generated, wiped and rebuilt on every Astro start |
 | `<pkg>/dist/` | `astro build` | the shipped site |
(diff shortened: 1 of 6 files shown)


codecalm:
Thanks for this, and especially for the writeup in #3006. Your diagnosis is right: `build-css.ts` calls `mkdirSync(outDir)` once at the top of `main()`, then spends a few seconds in sass before `flushWrites()` runs, and the `rmSync(publicDir, …)` in `copy-assets` lands somewhere in that window. Windows makes it reproducible because the recursive delete is slower there.

I'd rather fix it a level up, though. Calling `mkdirSync` before each write narrows the window but doesn't close it, and it leaves the actual problem alone: two processes both think they own `preview/public/`.

The rule we landed on a while back is that `copy-assets` is the only thing writing to `public/`, and everything else writes to `tmp-assets/` and gets synced in. We did that for `@tabler/docs` in 755bd889 and just forgot about `@tabler/preview`. So preview's `watch:css` now writes to `tmp-assets/css`, `syncDirs` picks it up, and `reloadDirs` is gone (nothing used it after that, and it was what encouraged writing into `public/` in the first place). `build-css.ts` doesn't need to change at all.

Nice side effect: docs already had `../preview/tmp-assets` in its `syncDirs`, but preview's watcher never wrote there, so docs never saw live `demo.css` edits. Now it does.

Tested from a cold start with the `public/` and `tmp-assets/` dirs removed: no `ENOENT`, `/preview/css/demo.css` serves 200, and editing `preview/scss/demo.scss` syncs through and reloads both dev servers.

I've pushed that to this branch since you'd enabled maintainer edits, so the net diff is now three files. Have a look and tell me if you'd rather I back it out. The diagnosis was the hard part here and that was yours.

ethancrawford:
All good, thanks for the correction!

PR #3070, opened 2026-09-16
Fix `.card-status-*` strip corner radius not matching the card

### LINKED ISSUE #3064: [BUG] Card status not correctly aligned
### Tabler version

1.5.1

### Browser

Firefox 155.0.1 (64 bit)

### OS

NixOS 26.11 (Zokor)

### Screen size

1920x1080

### Describe the bug

The card status is not correctly aligned because of border-radius.

### How to reproduce

1. Go to https://docs.tabler.io/ui/components/card#card-status
2. Zoom 
3. See your eyes bleed

### Screenshots

Image

### Live example

https://docs.tabler.io/ui/components/card#card-status

Bartosz-Do:
## Summary

- `.card-status-top`, `.card-status-start` and `.card-status-bottom` used to copy the card's own `border-radius`, but that radius gets clamped down by the strip's own thinness (it's only `border-width-wide`), so the strip's corner curved much tighter than the card's and left a visible gap or overhang at the corner — worse the bigger the card's radius.
- Each strip now rounds itself with its own pill radius, but only on the two corners on the side it sits on (the inner corners stay flush against the card body).
- The card gets a small matching radius (`card-status-size`) on that same side via `.card:has(> .card-status-*)`, so its edge lines up with the strip's curve instead of poking past it.

## Preview

- **URL:** [tabler-docs-git-fix-card-status-corner-radius-tabler-io.vercel.app/ui/components/card](https://tabler-docs-git-fix-card-status-corner-radius-tabler-io.vercel.app/ui/components/card#card-status)
- **How to test:** Open the "Card status" section (Top status / Bottom status / Side status). Zoom into a corner of any of the three cards — the colored strip's rounded end should follow the card's corner with no gap and no strip corner poking past the card. Same section is also live on [tabler-git-fix-card-status-corner-radius-tabler-io.vercel.app/cards.html](https://tabler-git-fix-card-status-corner-radius-tabler-io.vercel.app/cards.html) under "Cards inside card".

## Notes / rollout

- Closes #3064
- Changeset added for `@tabler/core` (patch).
- The new `.card:has()` rules pushed unminified `tabler.css` gzip size ~40 bytes past its 84 kB bundlewatch budget; raised to 85 kB in `core/package.json` (the minified bundle has plenty of headroom).


codecalm:
Thanks for picking this up, Bartosz. I pushed a different fix on top of your branch and wanted to explain why.

Your version made the card corners smaller whenever a card has a status strip (a 3px radius instead of the card's own 4px). That does make the two shapes match, but it changes how the card itself looks, and I'd rather not have a card change shape just because someone added a colored edge to it.

The root of the problem is that the strip was a separate 3px tall box with its own border-radius. The browser has to scale a 4px radius down to fit into a 3px box, so the strip's curve never lines up with the card's curve. You can't really fix that by picking a different radius for the strip.

So in 6f369bb the strip is now a full-size overlay on the card. It has the same border-radius as the card, a transparent 3px border on one side, and a mask (`mask-composite: exclude`) that cuts out everything except that border ring. The colored part is whatever is left, and it follows the card's corners exactly because it's literally the card's outline. Color still comes from the `bg-*` class, so the markup doesn't change. I added `pointer-events: none` so the overlay doesn't get in the way of clicks inside the card.

I checked it in Firefox at 6x zoom in both LTR and RTL, and the strip hugs the corner in all three positions. The SCSS ends up shorter than before, but the gzip size still goes up by a few bytes, and dev was already sitting right at the 84 kB budget. So I kept your bump to 85 kB and your changeset.

codecalm:
One more change in the follow-up commit. The `mask-composite: exclude` trick left a faint tint of the strip color on the two corners that don't have a border. That's an anti-aliasing thing: the border-box and padding-box masks don't cancel out perfectly along the curve where the border is 0px wide.

So the strip is now masked with a plain gradient instead, `linear-gradient(to bottom, #000 3px, transparent 3px)` for the top and the same idea for the other edges. The overlay still has the card's radius, so the outer edge of the strip follows the corner, and there's nothing left to bleed on the opposite side. The side strip gets a `[dir="rtl"]` override because rtlcss doesn't flip gradient directions inside `mask-image`. Checked again in Firefox at 1x and 6x, LTR and RTL.

PR #3096, opened 2026-09-18
Fix invisible text selection inside pre blocks

### LINKED ISSUE #3095: Text selection is barely visible inside `<pre>` blocks (light mode)
## Problem

The global `::selection` rule uses `$selection-bg`, which in light mode resolves to `color-mix(in srgb, var(--primary) 10%, transparent)`. That works on the light page background, but `<pre>` blocks are always rendered on `--bg-surface-dark`. A 10 % primary tint on a dark surface is almost invisible, so selected text in the "terminal"-style blocks (e.g. **Extra → Logs** in the preview) can hardly be distinguished from unselected text.

Screenshot from the preview, `Effective URL` is selected:

Selected text inside a pre block is barely visible

## Steps to reproduce

1. Open https://preview.tabler.io/logs.html in light mode.
2. Select some text inside one of the `<pre>` blocks.

## Suggested fix

`<pre>` keeps its dark background regardless of the colour scheme, so its selection colour should not follow `light-dark()` either. Applying the dark-mode value of `$selection-bg` (40 % primary) to `pre ::selection` gives a clearly visible highlight while staying within the existing palette:

```scss
$pre-selection-bg: color-transparent(var(--primary), 0.4) !default;

pre {
  ::selection { background-color: $pre-selection-bg; }
}
```

I can open a PR for this.

---
*This issue was drafted with the help of Claude (Anthropic).*

nerdoc:
Closes #3095

`<pre>` always renders on `--bg-surface-dark`, but the global `::selection` colour follows `light-dark()` and resolves to a 10 % primary tint in light mode - invisible on that background.

This adds `$pre-selection-bg` (40 % primary, the value dark mode already uses everywhere) and applies it to `pre::selection` / `pre ::selection`.

Before / after (light mode, "Effective URL" selected):

| before | after |
|---|---|
| before | after |

---
*Drafted with the help of Claude (Anthropic).*


codecalm:
Thanks for the fix and the clear write-up! I retargeted the PR to `dev`, which is our working branch. Merging.

nerdoc:
Ok, I'll try to memorize that, if I have another PR. And for the writeup, credits go to Claude ;-)
Everyone is eager to say how much "AI slop" ther is out in the wild. But AI really helps save time when you want to fix something and just let it do the paperwork.
Thanks for tabler.io.

PR #3111, opened 2026-09-21
Fix validation colors on `.input-group-text` in `.input-group-flat`

### LINKED ISSUE #1676: [BUG] Validation states in input-group-flat broken with input-group-text
**Describe the bug**

Looks broken with any `input-group-text` (gets worse when there is really text instead of an icon):
Bildschirmfoto 2023-08-18 um 16 39 18
Bildschirmfoto 2023-08-18 um 16 40 40

Looks good when I remove the `input-group-text` node:
Bildschirmfoto 2023-08-18 um 16 41 44

**Expected**

The `input-group-flat` rules should support `input-group-text` nodes in combination with validation states.

issue can be reproduced on the preview page by manipulating the HTML directly.

**Workaround**

Sharing a possible workaround for the next user with the problem:
Bildschirmfoto 2023-08-18 um 16 43 40

Use `is-invalid is-invalid-lite` ... but whether this is possible highly depends on the used framework.
I don't think this is a 100% viable solution, because it is visually rather hidden.

Bartosz-Do:
## Summary

- `.is-valid` and `.is-invalid` colored only the input in `.input-group-flat`. The `.input-group-text` kept a neutral border, so the field looked split in two.
- Focus made it worse. It replaced the red or green color of the whole group with the neutral focus color.
- `:has()` now passes the validation border and focus outline color to the addon. It works for `.is-valid`, `.is-invalid`, and for `.was-validated` with `:valid` and `:invalid`.
- Adds a test for the new selectors and a changeset.

## Screenshot

image

## Preview

- **URL:** [preview link](https://tabler-git-fix-gh-1676-input-group-flat-validation-tabler-io.vercel.app/form-elements.html)
- **How to test:** open the form elements page. In DevTools, add `is-invalid` to the input in "Input with prepended text". The whole field, including `https://`, should get a red border. Focus the input and check that the border and outline stay red. Repeat with `is-valid` for green. On `dev` the addon stays gray and focus turns the group blue.

## Notes / rollout

- Fixes #1676
- No demo page shows `input-group-flat` with a validation state, so the steps above use DevTools.
- Size: the bundlewatch check passes. `tabler.min.css` is 81.31 KB against the 82 KB limit, so there is little room left.
- Projects that worked around the bug with their own styles may now see the color applied twice.


kevinpapst:
Thanks for fixing @Bartosz-Do 👍 

In my opinion there should be an explicit demo section on the forms page for that, otherwise you / us cannot easily recognize if any change might break it again in the future. 

So for the sake of preventing regressions, I would vote for a small new box that covers all is-valid and is-invalid cases of the flat form.

PR #3124, opened 2026-09-21
Fix `tabler is not defined` on the Sparkline and Confetti pages

codecalm:
## Summary

- The Sparkline and Confetti preview pages came from `dev` and called the `tabler` global. 2.0 is ESM only, so there is no global and both pages threw `ReferenceError: tabler is not defined`.
- This made the `html-checks` job fail on `v2-dev` itself, and so on every open PR into `v2-dev`.
- The three page scripts are now `type="module"` and import `Sparkline` / `Confetti` from the `tabler.js` bundle.
- New `distJsPath()` helper in `shared/lib/assets.ts` gives pages and `BaseLayout` the same bundle url, so the browser does not evaluate the module twice.

## Preview

- **URL:** [sparkline](https://tabler-git-fix-v2-esm-demo-scripts-tabler-io.vercel.app/sparkline.html), [confetti](https://tabler-git-fix-v2-esm-demo-scripts-tabler-io.vercel.app/confetti.html)
- **How to test:** On the Sparkline page click "Randomize" in the "Update from code" card. The chart changes and the badge counts the updates. On the Confetti page a burst runs on load, and submitting the "Create" form runs another one. The console stays clean on both pages.

## Notes / rollout

- Checked locally the same way as CI: `check:html` reports 0 errors for `preview/dist` and `docs/dist`, and `check:markup-classes` passes.
- After the merge, open PRs into `v2-dev` need an update from the base to get a green `html-checks`.


codecalm:
Duplicate of #3105, which already landed in `v2-dev`.

